PR #7, opened 2021-08-04
feat: simulate body events

farnabaz:
This PR create a simple simulation for body streaming based on https://github.com/unjs/h3/issues/29#issuecomment-891951452

resolves #8


pi0:
(closing since timeout is not a good solution. since having workaround, let's redo it later and track in linked issue)

PR #20, opened 2021-11-19
fix: return `this` when calling `setHeader`, `end` and `destroy`

danielroe:
`setHeader` returns `this`

https://nodejs.org/api/http.html#responsesetheadername-value


pi0:
Nice find. Cheking code, it seems we also need to allow chaining for `end` and `destryo` (in wriable base) do you mind to also fix these two?

pi0:
@danielroe Can we maybe ignore return type (return this but type as void) until types are fixed?

Issue #21, opened 2021-12-09
`Writable["_data"]` can be null or undefined, with an unsafe `toString()`


tarantoj-intrepid:
In the following snippet, res._data has a type of `any`, meaning it can be `null | undefined`, which do not have `toString()` methods, causing a runtime error.
https://github.com/unjs/unenv/blob/fadb926e8827799aa0e5a33e5a5f0250f3003eac/src/runtime/fetch/call.ts#L35

Definition of the `_data` property is here
https://github.com/unjs/unenv/blob/fadb926e8827799aa0e5a33e5a5f0250f3003eac/src/runtime/node/stream/writable.ts#L19

`_data` should probably be typed better as either `_data: unknown` or `_data: string | number | object | array | undefined | null` to be able to catch this error at build

PR #22, opened 2021-12-09
fix: handle null or undefined _data

tarantoj-intrepid:
handle case where _data is null or undefined

Intended to address https://github.com/unjs/unenv/issues/21

### DIFF
--- a/src/runtime/fetch/call.ts
+++ b/src/runtime/fetch/call.ts
@@ -32,7 +32,7 @@ export function createCall (handle: Handle) {
 
     return handle(req, res).then(() => {
       const r = {
-        body: res._data.toString(),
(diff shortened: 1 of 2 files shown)


src/runtime/fetch/call.ts
```diff
@@ -32,7 +32,7 @@ export function createCall (handle: Handle) {
 
     return handle(req, res).then(() => {
       const r = {
-        body: res._data.toString(),
+        body: (res._data as any)?.toString() ?? "",
```
danielroe, line 35:
```suggestion
        body: (res._data as any)?.toString() ?? '',
```

danielroe (approved):

PR #43, opened 2022-06-29
feat: support returning raw body for binary content-types

danielroe:
related downstream issue: https://github.com/unjs/nitro/issues/100

### DIFF
--- a/src/runtime/fetch/call.ts
+++ b/src/runtime/fetch/call.ts
@@ -14,6 +14,11 @@ export interface CallContext {
   body?: any
 }
 
+const BINARY_CONTENT_TYPE_RE = /^(audio|font|video|image\/(?!svg)|application\/(?!json))/
+const isBinaryContentType = (contentType = '') =>
+  BINARY_CONTENT_TYPE_RE.test(contentType)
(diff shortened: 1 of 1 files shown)


src/runtime/fetch/call.ts
```diff
@@ -14,6 +14,8 @@ export interface CallContext {
   body?: any
 }
 
+const BINARY_CONTENT_TYPE_RE = /^(audio|font|video|image\/(?!svg)|application\/(?!json))/
```
pi0, line 17: Text mimes are more limited. We could inverse to list `text/*` + `application/json` as text content and assume the rest as binary. (maybe + empty content-type as text?)

mariocesar:
FYI, if this and https://github.com/unjs/nitro/pull/320 is merged it will enable to deploy Nuxt.js to AWS Lambda

pi0:
Let's return `_data` as is and let the upper layer (fetch) or direct use intercept the body type. It has to be a breaking change but seems better fix to not assume type of _data (technically for node polyfill it should be buffer or string but we set it as is there too)

PR #45, opened 2022-08-05
fix: order aliases from specific to general

danielroe:
Currently aliases are resolved from top down, so `stream/web` is (for example) resolved as `unenv/runtime/node/stream/index/web` rather than `unenv/runtime/node/stream/web/index`.

This PR reorders aliases so the more specific ones are at the top. Alternatively, we could use a helper function to do this for us.

https://github.com/unjs/nitro/issues/378

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -13,7 +13,7 @@ export function env (...presets: Preset[]): Environment {
     if (preset.alias) {
       // Sort aliases from specific to general (ie. fs/promises before fs)
       const aliases = Object.keys(preset.alias).sort((a, b) =>
-        b.split('/').length - a.split('/').length
+        (b.split('/').length - a.split('/').length) || (b.length - a.length)
       )
       for (const from of aliases) {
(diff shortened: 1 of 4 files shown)


pi0:
We might normalize and sort them in `env.ts#L18` once when generating preset.

src/env.ts
```diff
@@ -11,7 +11,11 @@ export function env (...presets: Preset[]): Environment {
   for (const preset of presets) {
     // Alias
     if (preset.alias) {
-      for (const from in preset.alias) {
+      // Sort aliases from specific to general (ie. fs/promises before fs)
+      const aliases = Object.keys(preset.alias).sort((a, b) =>
+        b.split('/').length - a.split('/').length
```
pi0, line 16: Right now we don't have prefix aliases but same as nitro, we could sort by length as fallback:

```suggestion
        (b.split('/').length - a.split('/').length) || (b.length-a.length)
```

PR #46, opened 2022-08-05
fix: allow mocking subpaths of proxy

danielroe:
https://github.com/nuxt/framework/issues/2724, https://github.com/nuxt/framework/issues/5482

When something that has been aliased to `proxy-cjs` is accessed with a subpath, it falls back to the next valid subpath, which will never succeed: import (`unenv/runtime/*` => `unenv/runtime/mock/proxy-cjs/.cjs`).


pi0 (approved):
Love the workaround!

PR #62, opened 2022-12-18
refactor: use type imports for unused node builtins

danielroe:


pi0 (approved):
Thanks!

pi0:
We might enable eslint rule as well

PR #63, opened 2022-12-31
fix: export named `fetch` from `node-fetch` polyfill

danielroe:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)

### 📚 Description

This adds a named fetch export for compatibility with `node-fetch-native`: https://github.com/unjs/node-fetch-native/blob/main/src/index.ts#L12.

### 📝 Checklist

### DIFF
--- a/src/runtime/npm/node-fetch.ts
+++ b/src/runtime/npm/node-fetch.ts
@@ -1,7 +1,7 @@
 // https://github.com/node-fetch/node-fetch
 
 // Native browser APIs
-const fetch = (...args: Parameters<typeof globalThis["fetch"]>) => globalThis.fetch(...args);
+export const fetch = (...args: Parameters<typeof globalThis["fetch"]>) => globalThis.fetch(...args);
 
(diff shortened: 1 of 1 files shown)


pi0 (approved):

pi0:
Seems a nice addition btw we are not aliasing `node-fetch-native` itself ([src](https://github.com/unjs/unenv/blob/4dcf17d574d370af43bc919582f1e380d1aa94cb/src/presets/node.ts#L5)). Wondering what was the original issue?

rvmourik:
@danielroe do you know when there will be a tagged release with this commit? We encounter this error and it stops us from deploying our newly migrated Nuxt3 website to Cloudflare workers.

Thanks in advance 🙏

danielroe:
You should now be able to benefit from this from `v1.0.1` ❤️

rvmourik:
> You should now be able to benefit from this from `v1.0.1` ❤️

Thanks, works perfectly.

PR #100, opened 2023-04-21
feat(node/crypto): implement `randomBytes`

BoxenOfDonuts:
Added export of `randomBytes` to the crypto polyfill.  If another dependency requires it currently, nitro will fail to build. Currently webcrypto does not have a `randomBytes` method

```
[nitro 2:36:51 PM]  ERROR  RollupError: "randomBytes" is not exported by "node_modules/unenv/runtime/node/crypto/index.mjs", imported by "node_modules/@firebase/firestore/dist/lite/index.node.mjs".


5: import { FirebaseError, getDefaultEmulatorHostnameAndPort, createMockUserToken, getModularInstance, deepEqual } from ...
6: import nodeFetch from 'node-fetch';
7: import { randomBytes as randomBytes$1 } from 'crypto';
            ^
8:
9: const version$1 = "3.7.2";


 ERROR  "randomBytes" is not exported by "node_modules/unenv/runtime/node/crypto/index.mjs", imported by "node_modules/@firebase/firestore/dist/lite/index.node.mjs".
```

### DIFF
--- a/src/runtime/node/crypto/index.ts
+++ b/src/runtime/node/crypto/index.ts
@@ -1,6 +1,7 @@
 // https://nodejs.org/api/crypto.html
 // https://github.com/unjs/uncrypto
 import type nodeCrypto from "node:crypto";
+import { notImplemented } from "src/runtime/_internal/utils";
 
 const webcrypto = globalThis.crypto;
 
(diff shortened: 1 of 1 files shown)


pi0:
Thanks. This looks good. We can also polyfill it ([like this](https://github.com/browserify/randombytes/blob/master/browser.js)). Would you like to help on this in same PR?

BoxenOfDonuts:
Yes, I can give it a shot

BoxenOfDonuts:
Removed some guard rails, let me know if you want them back in.

Removed:
1.  Checking `MAX_UINT32` as the internal Buffer will throw already. 
2. The If check for `Size > 0`
3. The If check to either slice or return `getRandomValues` since slicing longer than the TypedArray should just return the TypedArray

pi0 (approved):
Thanks! 

Made few changes to suppress type issues, use global Buffer and Uinit8Array super class

Issue #110, opened 2023-06-28
cloudflare-pages uses crypto polyfill but cant find methods


gritworks:
### Describe the feature

I am creating a nuxt3 app and want to swap vercel for cloudflare-pages.

Currently its deployed on vercel, not sure why the (probably incomplete?) crypto polyfill provided by unenv is no issue here. The app is running fine. of course the preset for nitro is different.

On cloudflare-pages there seems to be an issue with libraries depending on unenv 

- jose
- nitropack
- sidebase/nuxt-auth

as someone mentioned it was worth trying to override unenv for these packages above to be of version 1.2.2 (in package.json)

this wont work either as it looks like jose is depending on newer exports from unenv. ("types" is not exported from util polyfill by unenv but wanted by jose) so the util library is a core node library that seems to be polyfilled incompletely also. Correct me if i'm wrong as i'm no expert at all when it comes to node.

I further looked at the possibility to maybe inject missing functions with the vite bundler but it seems only to be possible with core library's.

Hope this makes any sense

Kind regards,
Wouter

### Additional information

- [ ] Would you be willing to help implement this feature?

AwesomeDude091:
I think it’s because nitro compiles it into a different format. https://nitro.unjs.io/deploy/providers/cloudflare 

Also I can confirm this is an issue

AwesomeDude091:
The above Pull Request allows for sidebase/nuxt-auth to compile

Hebilicious:
Hi @AwesomeDude091, this error isn't necessarily related to unenv, but to the fact that sidebase/nuxt-auth isn't compatible with cloudflare.

You could use this module I made for my personal needs https://github.com/Hebilicious/authjs-nuxt

What's happening is that despite the fact that some of nuxt-auth dependencies use the "node:crypto" module, which isn't a node only API. While you could try to polyfill it with unenv, you might run into other issues. The proper fix is to use packages that uses "crypto" ("web:crypto" in node.js) if you're not using node.js environments.

PR #111, opened 2023-06-28
feat: support more node crypto functions

### LINKED ISSUE #91: Complete node specific `crypto` polyfills
(followup for #90)

AwesomeDude091:
Hopefully a fix for #110 
Resolves #91

### DIFF
--- a/src/runtime/node/crypto/index.ts
+++ b/src/runtime/node/crypto/index.ts
@@ -1,8 +1,11 @@
 // https://nodejs.org/api/crypto.html
 // https://github.com/unjs/uncrypto
 import nodeCrypto from "node:crypto";
+import stream from "node:stream";
 
 const webcrypto = globalThis.crypto;
+
+export const constants = nodeCrypto.constants
 
 export const subtle: typeof nodeCrypto.subtle = webcrypto.subtle;
 
@@ -19,20 +22,6 @@ export const getRandomValues: typeof nodeCrypto.getRandomValues = (
 ) => {
   return webcrypto.getRandomValues(array);
 };
-
-export const createHash: typeof nodeCrypto.createHash = (
-    algorithm,
-    options?
-) => {
-  return nodeCrypto.createHash(algorithm, options)
-}
-
-export const createHmac: typeof nodeCrypto.createHmac = (
-    algorithm,
-    options?
-) => {
-  return nodeCrypto.createHmac(algorithm, options)
-}
 
 export const randomBytes /* :typeof nodeCrypto.randomBytes */ = (
   size: number,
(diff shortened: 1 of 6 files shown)


AwesomeDude091:
Are there any more changes required @pi0 ?

pi0:
Thanks for the pull request dear @AwesomeDude091. This PR is great however we probably need to use more of webCrypto API as node:crypto is only available for Node.js environments and compatibility layer of cloudflare pages for node.

Hebilicious:
@AwesomeDude091 @Romitou I would be surprised if this solves your issues with cloudflare *once it's deployed*.
Building locally will work because you are using a node.js environment for the build, however, what you're doing here isn't a polyfill, you're just using "node:crypto" in unenv (which should break in workerd as "node:crypto" won't be available).

If you want to polyfill all "node:crypto" method, you would basically need to do a real polyfill of node:crypto, similar to https://github.com/browserify/crypto-browserify 

What I can suggest instead is that instead of trying to polyfill `nuxt-auth` with cloudflare, you use another library that doesn't require node-only internals at runtime, such as https://github.com/Hebilicious/authjs-nuxt or Lucia https://lucia-auth.com/start-here/getting-started?nuxt or even Hanko https://github.com/nuxt-modules/hanko

Hebilicious (changes requested):
You shouldn't import any node internals in the runtime. 
Type imports are fine.

Hebilicious:
@pi0 we could go with an incremental approach for the crypto polyfills, starting with the functions needed for nuxt-auth seems less daunting.
We also need to make sure that we don't polyfill things that are available in crypto global

AwesomeDude091:
Update nearly all Cloudflare Issues have been resolved with the current build (testing on a local wrangler). It builds successfully and without warnings, however the worker gets fails on initalization with error ` Uncaught TypeError: Cannot read properties of undefined (reading 'split')
  at do835gcosb7.js:20707:33`, however this seems to be more of a general issue with cloudflare and nuxt also outlined here:  https://github.com/cloudflare/workers-sdk/issues/3144
  
Any ideas on how to tackle that 'split' issue?  
  
P.S. For testing I build unenv locally paste it into my node_modules folder of my test application (overwritting the current released build) and then build and wrangler execute

AwesomeDude091:
I isolated the line that is causing issues in the bundledWorker:
`var [Bv, Mv] = ge.versions.node.split(".").map((F2) => parseInt(F2, 10));`

Hebilicious:
> I isolated the line that is causing issues in the bundledWorker: `var [Bv, Mv] = ge.versions.node.split(".").map((F2) => parseInt(F2, 10));`
> 
> ```
> var me = "undefined" != typeof globalThis ? globalThis : "undefined" != typeof self ? self : "undefined" != typeof global ? global : {};
> me.process = me.process || oe;
> var ge = me.process;
> ```
> 
> Do you guys @pi0 @Hebilicious have any ideas?

Read my comment above about using node internals in unenv. This approach won't work, you need to remove all imports from "node:crypto" and *polyfill* them. You can keep the type imports as they won't be included in the final bundle.

AwesomeDude091:
I need some help, here it is not working out

AwesomeDude091:
I cannot figure out the port for generateKey, generateKeySync, generateKeyPair, generateKeyPairSync

AwesomeDude091:
> I cannot figure out the port for generateKey, generateKeySync, generateKeyPair, generateKeyPairSync

Any help with this would be much appreciated

AwesomeDude091:
TODOs for a nearly complete Crypto Polyfill

- [ ] generateKey
- [ ] generateKeySync
- [ ] generateKeyPair
- [ ] generateKeyPairSync
- [ ] KeyObject

AwesomeDude091:
@pi0 @Hebilicious Does the above list check out, are the previous new commits okay and is there any chance that you could give me a pointer as to what to do for the generateKey and generateKeyPair logic, especially the difference between sync and not sync?

Hebilicious (changes requested):
Nice work @AwesomeDude091 !

You are attempting to do something that is extremely complicated. Supporting ALL of node crypto internals might be unachievable. 

Even if you succeed, this approach has some downsides : 

- Runtime size : Workers environments are designed to run with low resources. Even if you can implement the more complex crypto functionalities, they are CPU heavy and you might not be able to use them in a worker.

- Security : Poyfill these things could open your applications to vulnerability as you are now in charge of making it safe.

I'm not a fan of adding browserify as a dependency. For our use case, it might be better to copy their logic instead of using their package.

What I would like to make these extra polyfill opt-in, ie have them behind some kind of configuration flag such as `preset.extraCryptoPolyfill` ...

If you *still* wish to continue working on this, I can recommend you take a look at https://github.com/PeculiarVentures/webcrypto and their other repositories.

As I said before, I wouldn't do all of this *just* to use`sidebase/nuxt-auth` on cloudflare workers. I would only recommend you do this if you're really interested in cryptography and to find out how to do this, since you're in un-charted territory and it will be relatively hard to help you.

Bonus reading : 
https://tonyarcieri.com/whats-wrong-with-webcrypto

pi0:
I will be working on this PR/Issue later this week once had time. 

In general overview for changes:

- Partially reimplementing Node API for web, is WHAT UNENV IS BULT FOR so this PR is not in a wrong direction 👍🏼 
- We shall not import directly from `node:` namespace (unless for types) and implement from `unenv/node/stream` for example for stream utils when we need
- When cloudflare and other runtimes have more stable polyfill, we shall use it (it is unrelated to this PR btw and more about higher order usage in Nitro)

src/runtime/node/util/types/_types.ts
```diff
@@ -165,12 +166,10 @@ export const isBigInt64Array: typeof utilTypes.isBigInt64Array = notImplemented(
 export const isBigUint64Array: typeof utilTypes.isBigUint64Array =
   notImplemented("util.types.isBigUint64Array");
 
-// @ts-expect-error not implemented yet
-export const isKeyObject: typeof utilTypes.isKeyObject = notImplemented(
-  "util.types.isKeyObject"
-);
+// @ts-ignore
```
Hebilicious, line 169: should be ts-expect-error

src/runtime/node/crypto/index.ts
```diff
@@ -1,8 +1,17 @@
 // https://nodejs.org/api/crypto.html
 // https://github.com/unjs/uncrypto
-import type nodeCrypto from "node:crypto";
+import type nodeCrypto from "node:crypto"
+import stream from "node:stream";
+// @ts-ignore
+import cryptoBrowserify from "crypto-browserify"
+// @ts-ignore
+import { webTimingSafeEqual } from "@advena/web-timing-safe-equal"
```
Hebilicious, line 8: we should use the browserify version instead of this one https://www.npmjs.com/package/timing-safe-equal

Hebilicious (changes requested):
- Some comments that needs to be removed/included.
- Should we proxy all of browserify-crypto and augment it with the correct types ?
- Should unenv have an explicit opt-in behaviour for these browserify crypto polyfills ? This would be consistent with something like https://github.com/cyco130/esbuild-plugin-polyfill-node
- What happens if only one import is needed ? Does *everything* gets added to the final bundle ? @pi0

Hebilicious (changes requested):
duplicate

pi0:
@Hebilicious Thanks for the review. I will handle the next steps on this. Kindly please NEVER request for change in other PRs in unjs repos unless you are an explicit maintainer of it ([read more about roles](https://github.com/unjs/governance/#unjs-team)). You can always ask questions about general project directions from maintainers.

> Should we proxy all of browserify-crypto and augment it with the correct types ?

Not necessarily for all. Only for ones are possible to be purely implemented in JS with Web APIs. For those that are not possible or not implemented yet, unenv simply throws an `Not Implemented` error. But this often helps to fix lots of build errors with incompatible Node.js libraries that simply import things and in practice don't essentially require them.

> Should unenv have an explicit opt-in behavior for these browserify crypto polyfills ? This would be consistent with something like https://github.com/cyco130/esbuild-plugin-polyfill-node

Unenv only provides presets and polyfill (today). It is up to the consumer (nitro cloudflare presets for instance) to apply opt-in/out behavior which has to be opt-out i guess since cloudflare (and only with a flag) allows some of Node apis.

> What happens if only one import is needed ? Does everything gets added to the final bundle ? 

Exports are tree-shakable in general unless a lib is using `require` with untreeshakable syntax or `import * from`.

AwesomeDude091:
@pi0 Are there TransformOptions / StreamOptions implemenations as part of the UNENV polyfill?

AwesomeDude091:
My bad, figured it out

AwesomeDude091:
@pi0 I have started the implementation of  `generateKeyPair` hope this is the right direction...

pi0:
Hi dear @AwesomeDude091 thanks again for working on all this and so sorry for all troubles. 

#121 adds all needed stubs so builds won't fail. 

For cloudflare (and other worker runtimes) we shall leverage [Node.js compatibility](https://developers.cloudflare.com/workers/runtime-apis/nodejs/)

If you think some of utils could be actually implemented and possible without external dependencies such as browserify-crypto and purely based on webcrypto/subtle, PR more than welcome on case by case basis.

Best luck ❤️

PR #122, opened 2023-08-07
fix(node:crypto): properly call web crypto methods

AaronDewes:
The latest change broke some Nuxt deployments to Deno deploy for me. Deno's webcrypto is somewhat different, and `...globalThis.crypto` does not work in Deno. This replaces it with a custom getter that retrieves the correct value instead. This adds some overhead, but is required for Deno compatibility.

grafik

### DIFF
--- a/src/runtime/node/crypto/web.ts
+++ b/src/runtime/node/crypto/web.ts
@@ -5,10 +5,16 @@ import type nodeCrypto from "node:crypto";
 export const CryptoKey =
   globalThis.CryptoKey as unknown as typeof nodeCrypto.webcrypto.CryptoKey;
 
-export const webcrypto: Crypto & typeof nodeCrypto.webcrypto = {
-  CryptoKey,
-  ...globalThis.crypto,
-};
+function ensureBound(thisArg: any, name: string | symbol) {
+  const fn = thisArg[name];
+  return typeof fn === "function" ? fn.bind(thisArg) : fn;
+}
+
+export const webcrypto: Crypto & typeof nodeCrypto.webcrypto = new Proxy<Crypto & typeof nodeCrypto.webcrypto>({} as any, {
+  get(target, name) {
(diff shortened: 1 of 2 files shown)


src/runtime/node/crypto/web.ts
```diff
@@ -5,10 +5,16 @@ import type nodeCrypto from "node:crypto";
 export const CryptoKey =
   globalThis.CryptoKey as unknown as typeof nodeCrypto.webcrypto.CryptoKey;
 
-export const webcrypto: Crypto & typeof nodeCrypto.webcrypto = {
-  CryptoKey,
```
pi0, line 9: Maybe we can simply bind CryptoKey in L5 to avoid proxy?

AaronDewes: I haven't tried that, and I don't think I fully understand what you mean to be honest. Can you show me a code example so I can try locally?

pi0: `const CryptoKey = globalThis.CryptoKey.bind(globalThis.CryptoKey)`. Please try it on your repo (if you have a reproduction that would be amazing too!)

AaronDewes: I'm using Nuxt with Deno Deploy, so the issue is somewhat nested in the dependency chain, but I'll try to create a minimal sample.

AaronDewes: https://github.com/AaronDewes/unenv-bug-demo There's also some more explanation of the bug in this repo. I hope this shows the issue well enough. Nitro is probably not needed, but was the easiest for me to set up as a demo.

pi0: Thank you for nice reproduction! Have made some few refactors to make sure node webcrypto is not affecting us also simplified proxy (by using main crypto as target, proxy handles calls with the right context)

Issue #72, opened 2023-01-28
Make the platform's context available


dario-piotrowicz:
### Describe the feature

Hi, I'm trying to create a Nuxt application and deploy it via [Cloudflare Pages](https://pages.cloudflare.com/)

When you deploy an application via Cloudflare Pages you can access other Cloudflare products via [bindings](https://developers.cloudflare.com/pages/platform/functions/bindings/) (which basically allow you to store data, etc...)

The issue is that I believe that such bindings are not made accessible via H3 (I expected to maybe find them in the `H3EventContext` but no luck, as you can see here: [my gh nuxt get context test](https://github.com/dario-piotrowicz/my-nuxt-app/blob/88f115dc2b99caf4b4e64403c2400314ebc92fce/server/api/test.get.ts) (deployed [here](https://my-nuxt-app-edm.pages.dev/)))

Edge Frameworks which I am used to all make such context available, like for example:
 - [Itty router](https://itty-router.dev/#/#cloudflare-es6-module-syntax-required-for-durable-objects-a-idcf-es6-module-syntaxa)
 - [Hono](https://honojs.dev/docs/api/context/#cenv)
 - [Remix](https://remix.run/docs/en/v1/guides/data-loading#cloudflare-kv)
 - [SvelteKit](https://kit.svelte.dev/docs/adapter-cloudflare#environment-variables)
 - [Qwik](https://qwik.builder.io/integrations/deployments/cloudflare-pages/#context)
 - etc...
 
 Would there be any chance for this context to also be provided by H3? :pray: 
 (Because otherwise this creates a non trivial limitation to the use of Nuxt with Cloudflare Pages)
 
 (Naturally please let me know if the context is actually available somewhere and I just missed it :sweat_smile: )

### Additional information

- [X] Would you be willing to help implement this feature?

pi0:
A main blocker was that we were not able to pass a context to direct calls. It is now supported by https://github.com/unjs/unenv/commit/a9190c4120aa3ab2edcbb77f780f9e3f60b47c55 in latest version of unenv.

dario-piotrowicz:
@pi0 the context being passed doesn't seem the one needed for the bindings as it only contains the `cf` property

<details>
<summary>For example</summary>

This is what I get as the `__unenv__`:
```
{
  "cf": {
      "clientTcpRtt": 7,
      "longitude": "-0.02000",
      "latitude": "51.50640",
      "tlsCipher": "AEAD-AES128-GCM-SHA256",
[...]

PR #125, opened 2023-08-08
feat: expose `__unenv__` flag for unenv classes and functions

pi0:
### ❓ Type of change

### 📚 Description

it is commonly required to identify if something is coming from unenv or not specially for classes and mocked objects. 

This PR adds `__unenv__` property to class (instances) and functions to identify.

**Note:** Since we were already using `req.__unenv__` for another purpose (https://github.com/unjs/unenv/issues/72), the type for `IncommingMessage` and `Readable` classes is unknown/object.

### 📝 Checklist

### DIFF
--- a/src/runtime/mock/empty.ts
+++ b/src/runtime/mock/empty.ts
@@ -1,1 +1,5 @@
-export default {};
+export default Object.freeze(
+  Object.create(null, {
+    __unenv__: { get: () => true },
+  }),
+);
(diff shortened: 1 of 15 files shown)


src/runtime/mock/empty.ts
```diff
@@ -1 +1,5 @@
-export default {};
+export default Object.freeze(
```
pi0, line 1: Note: freezing is not related to this PR but quickly included as fix

PR #168, opened 2024-03-18
feat(node:process): add polyfill for `process.hrtime`

alan-agius4:
### ❓ Type of change

- [] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

`process.hrtime` is currently not implemented.

### 📝 Checklist

### DIFF
--- a/src/runtime/node/process/_process.ts
+++ b/src/runtime/node/process/_process.ts
@@ -226,3 +226,12 @@ process.chdir = function (dir) {
 process.umask = function () {
   return 0;
 };
+
+// https://nodejs.org/api/process.html#processhrtime
+process.hrtime = function () {
+  return [0, 0];
+};
+process.hrtime.bigint = function () {
+  // Convert milliseconds to nanoseconds
+  return BigInt(Date.now() * 1_000_000);
+};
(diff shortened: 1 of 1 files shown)


src/runtime/node/process/_process.ts
```diff
@@ -226,3 +226,12 @@ process.chdir = function (dir) {
 process.umask = function () {
   return 0;
 };
+
+// https://nodejs.org/api/process.html#processhrtime
+process.hrtime = function () {
+  return [0, 0];
+};
```
IgorMinar, line 233: hey @alan-agius4, I think you actually want something closer to this:

```suggestion
process.hrtime = function (startTime) {
  const now = Date.now();
  // millis to seconds
  const seconds = Math.trunc(now / 1_000);
  // convert millis to nanos
  const nanos = (now % 1_000) * 1_000_000;
  
  if (startTime) {
    let diffSeconds = seconds - startTime[0];
    let diffNanos = nanos - startTime[0];
    
    if (diffNanos < 0) {
      diffSeconds = diffSeconds - 1;
      diffNanos = 1_000_000_000 + diffNanos;
    }
    return [diffSeconds, diffNanos]
  }
  
  return [seconds, nanos];
};
```

IgorMinar (commented):
This does look like a good change to me though, especially if the hrtime polyfill is tweaked.

pi0 (approved):
Thank you both ❤️

src/runtime/node/process/_process.ts
alan-agius4: Thanks @IgorMinar 😀

Issue #182, opened 2024-05-14
Adopt `process.getBuiltinModule(id)` for node polyfills


pi0:
(context: found about it from https://github.com/unjs/unenv/issues/181 thanks to @IgorMinar)

The upcoming [`process.getBuiltinModule(id)`](https://github.com/nodejs/node/pull/52762) API coming to Node.js allows much better adoption to partially have Node.js polyfills.

Issue #181, opened 2024-05-14
Support for hybrid (node) module polyfills


IgorMinar:
### Describe the feature

Unenv's presets allow us to mark certain modules as natively supported by a given host environment. For example Cloudflare's workerd natively supports several APIs in `node:buffer`, and the cloudflare preset specifies this module as natively supported: https://github.com/unjs/unenv/blob/c6dca1dfac95bd6359e8575d4456635914823701/src/presets/cloudflare.ts#L8

The issue I see is that workerd supports only some `node:buffer` APIs but not all. Specifically it supports `Buffer` and `SlowBuffer` APIs, but not `Blob` or `File`, which are part of Node's `buffer` module.

What I'd really like is if I could modify the Cloudflare preset in a way that would create a hybrid module polyfill for `node:buffer` which would provide the natively supported APIs of workerd's `node:module` while, using the unenv mock (or polyfills` for the remaining APIs.

I gave this a shot locally, but I wonder if this is actually possible at all with the current way unenv works. The problem I see is that we'd need a way to preserve access to the "native" `node:buffer` while also defining an `alias` for it. And since all aliases are flattened, I don't see a way how we can have both. Am I missing something? Is there some creative solution to this?

Or is the only way to achieve this to modify workerd to alias `node:*` modules via an alternative module prefix (e.g. `node-unenv:*`) so that unenv can internally use these modules while overriding `node:*` specifiers using module aliases?

### Additional information

- [x] Would you be willing to help implement this feature?

pi0:
It seems an interesting idea. If workerd could provide an alternative way to access node polyfills like `workerd:node:buffer` it would make it much easier since we could partially import them in our polyfills (even opened a better way to opt-in to things like async context)

An alternative would involved unenv + specific bundler (rollup/vite/nitro) plugin that preserves external imports from `node_modules/unenv/cloudflare.mjs`. It is certainly doable but also would be not most stable path because we need to also allow `node:` externals in `renderChunk` hook and in there we don't have a way to tell if source of `node:buffer` import was the unenv polyfill parent or a virtual file for example.

IgorMinar:
ok, thanks for the confirmation @pi0. I'll discuss with the team, and get back to you.

IgorMinar:
We discussed this and it appears that the easiest way for us to support this is via a newly proposed (and almost landed) [process.getBuiltinModule](https://github.com/nodejs/node/pull/52762) API.

[...]

PR #183, opened 2024-05-15
fix(cloudflare): make sure not fallback to `globalThis` for env

jculvey:
### ❓ Type of change

### 📝 Checklist

Adds a polyfill for `process` to the Cloudflare preset which combines the [node polyfill](https://github.com/unjs/unenv/blob/main/src/runtime/node/process/_process.ts) with the [workerd polyfill](https://github.com/cloudflare/workerd/blob/main/src/node/internal/process.ts) implementation. We'd specifically like to retain the polyfilled `nextTick` from workerd which uses `queueMicrotask`, but benefit from all the other global `process` polyfills provided by unenv.

### DIFF
--- a/src/runtime/polyfill/cloudflare.ts
+++ b/src/runtime/polyfill/cloudflare.ts
@@ -1,1 +1,2 @@
+// Make sure env polyfill won't fallback to legacy `globalThis` used for service-worker format
 (globalThis as any).__env__ = {};
--- a/src/runtime/polyfill/cloudflare.ts
+++ b/src/runtime/polyfill/cloudflare.ts
@@ -1,6 +1,1 @@
-type GlobalThis = typeof globalThis;
-interface GlobalThisPlusEnv extends GlobalThis {
-  __env__: typeof process.env;
-}
-
-(globalThis as GlobalThisPlusEnv).__env__ = {};
+(globalThis as any).__env__ = {};
--- a/src/runtime/polyfill/cloudflare.ts
+++ b/src/runtime/polyfill/cloudflare.ts
@@ -1,4 +1,6 @@
-import { nextTick } from "node:process";
-import { process as _process } from "../node/process/_process";
+type GlobalThis = typeof globalThis;
+interface GlobalThisPlusEnv extends GlobalThis {
+  __env__: typeof process.env;
+}
 
-Object.assign(_process, { nextTick });
+(globalThis as GlobalThisPlusEnv).__env__ = {};
--- a/src/runtime/polyfill/cloudflare.ts
+++ b/src/runtime/polyfill/cloudflare.ts
@@ -1,0 +1,4 @@
+import { nextTick } from "node:process";
+import { process as _process } from "../node/process/_process";
+
+Object.assign(_process, { nextTick });
--- a/src/presets/cloudflare.ts
+++ b/src/presets/cloudflare.ts
@@ -35,7 +35,7 @@ const cloudflarePreset: Preset = {
     ),
   },
   inject: {},
-  polyfill: [],
+  polyfill: ["unenv/runtime/polyfill/cloudflare"],
   external: cloudflareNodeCompatModules.map((p) => `node:${p}`),
 };
 


pi0:
Can you please update the polyfill to explictly override nextTick impl? 

(env implementation is more generic in unenv. we can explicitly add any more override in the future that needed) 

Also would it make sense to have a generic polyfills/cloudflare file that adds more of future polyfills via one entry?

jculvey:
> Can you please update the polyfill to explictly override nextTick impl?
> 
> (env implementation is more generic in unenv. we can explicitly add any more override in the future that needed)
> 
> Also would it make sense to have a generic polyfills/cloudflare file that adds more of future polyfills via one entry?

Great suggestions, updated!

Thanks for the speedy response time 😄.

src/runtime/polyfill/cloudflare.ts
```diff
@@ -0,0 +1,4 @@
+import { nextTick } from "node:process";
+import { process as _process } from "../node/process/_process";
+
+Object.assign(_process, { nextTick });
```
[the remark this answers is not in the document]
pi0: And one more! We could also _prefer_ (if available) `queueMicrotask` in [main process polyfill](https://github.com/unjs/unenv/blob/main/src/runtime/node/process/_process.ts#L137) with an early return. This makes the fix generic for all compatible runtimes.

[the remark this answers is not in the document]
IgorMinar: > And one more! We could also prefer (if available) queueMicrotask in main process polyfill with an early return. This makes the fix generic for all compatible runtimes.

That polyfill relies on setTimeout to schedule the work which is not the same as `queueMicrotask` — microtasks take precedence over tasks, so the current polyfill will run the callbacks in incorrect order.

Is there any reason for the nextTick polyfill not to use queueMicrotask? It seems to be widely supported API: https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask

> would it make sense that we avoid node: imports or do you see that in the future workerd might do it differently?

We do expect the workerd runtime to get a few more (process) APIs polyfilled over time, but more importantly this PR is an attempt to create a hybrid polyfill at a symbol level, similarly how https://github.com/unjs/unenv/issues/181 proposes it for the module level. We want to prove that this is possible, as that would then inform our overall node compat strategy.

src/runtime/polyfill/cloudflare.ts
```diff
@@ -0,0 +1,4 @@
+import { nextTick } from "node:process";
+import { process as _process } from "../node/process/_process";
```
IgorMinar, line 2: can you add a comment here about the issue with importing `../node/process/import` directly? it might be tempting for someone to refactor this in the future and run into the issues with messed up bundling we saw and can't explain.

src/runtime/polyfill/cloudflare.ts
```diff
@@ -0,0 +1,4 @@
+import { nextTick } from "node:process";
```
IgorMinar, line 1: didn't you want to import the whole module and spread it? or would that introduce some additional properties on process that should not be on the global? (e.g. default`).

IgorMinar: ah.. I see that @pi0 requested this change. I think we could keep as is then, and we can reconsider once there are more keys we'd like to use from the native layer.

However, I'm not quite sure about the `process.env` implementation in unenv: https://github.com/unjs/unenv/blob/bf79a17667cc8c8f0794800e1a50dea7e4c967cc/src/runtime/node/process/_process.ts#L160-L187

It comes across to me as very convoluted and potentially incorrect as reads and write target different underlying storage. @pi0 can you shed some light on why the env polyfill works this way? it might be the best to add a comment into that code to explain the reasoning.

src/runtime/polyfill/cloudflare.ts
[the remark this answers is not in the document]
pi0: No reason main polyfill not using `queueMicrotask` it was simply remaining from the legacy polyfills I used. 

We can use a condition in there to use `queueMicrotask` when available and eventually drop fallback if most of the runtimes support it.

> We want to prove that this is possible, as that would then inform our overall node compat strategy.

If it is helpful for your testing strategy, looks good to me this way 👍🏼 

From https://github.com/unjs/unenv/issues/181#issuecomment-2110447976 I was under the impression that you would want to go with `process.getBuiltinModule` API and that this PR particularly can be natively improved in main unenv's `nextTick` polyfill.

src/runtime/polyfill/cloudflare.ts
pi0: This allows Nitro users to access variables using `process.env[key]` (this is not normally possible via cloudflare edge)

I would be happy to explain more later if you wanted and see how to improve it :)

src/runtime/polyfill/cloudflare.ts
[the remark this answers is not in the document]
jculvey: @pi0 

I ended up importing `node:process` this way because I couldn't find another way to grab the native workerd global `process` in a reliable way. Is there a better way of doing it?

These all didn't seem to work:

```ts
console.log(process); //undefined
console.log(globalThis.process); // undefined
console.log(global.process); // undefined
```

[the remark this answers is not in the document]
IgorMinar: @jculvey the issue here is that workerd doesn't expose the process object on global at all (unless you use the underutilized `nodeJsCompatModule`, which we want to avoid because we are thinking of removing it).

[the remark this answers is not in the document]
IgorMinar: globalThis.process and `node:process` seem to have the same shape however, with the exception of `node:process#default` which exists to support default import.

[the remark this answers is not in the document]
jculvey: Right, my mistake. Thanks!

src/runtime/polyfill/cloudflare.ts
IgorMinar: I see. I think I understand how nitro uses`globalThis.__env__` and I don't mind that part, but what I find problematic is the fallback on `globalThis` which is used as the fallback here:

https://github.com/unjs/unenv/blob/bf79a17667cc8c8f0794800e1a50dea7e4c967cc/src/runtime/node/process/_process.ts#L163C41-L163C72

If I'm reading the code right, then n an environment like workerd, we'd end up reading `globalThis.foo` when someone tries to access `globalThis.process.env.foo` - this is not desirable for us.

I'm also puzzled about these two lines in `get` and `has`:
https://github.com/unjs/unenv/blob/bf79a17667cc8c8f0794800e1a50dea7e4c967cc/src/runtime/node/process/_process.ts#L168
https://github.com/unjs/unenv/blob/bf79a17667cc8c8f0794800e1a50dea7e4c967cc/src/runtime/node/process/_process.ts#L172

Since these two lines prefer `env` over `_envShim`, this means that any assignments or deletions performed on process.env will not be observable/readable if the key is present on the original `env`. This seems like a bug to me.

pi0: > If I'm reading the code right, then in an environment like workerd, we'd end up reading globalThis.foo when someone tries to access globalThis.process.env.foo - this is not desirable for us.

Indeed and it was because of cloudlare worker format support actually. As far as I remember it was a requirement for the cloudflare (legacy) worker format that bindings were accessible via `globalThis`. 

I am up to avoid this and probably workaround in nitro legacy worker preset but it needs a major version of `unenv` or at least waiting for a minor version of Nitro to allow this.

IgorMinar:
> On this PR, I would suggest to introduce runtime/polyfill/cloudflare.ts mainly to set __env__ shim override (to solve glboalThis concern of process.env https://github.com/unjs/unenv/pull/183#discussion_r1602104944) and wait for https://github.com/unjs/unenv/issues/182 to be landing before trying the mixed method.

I addressed this in https://github.com/unjs/unenv/pull/183#discussion_r1602197176

> Using node:process import idea, is probably not going to work well in practice because either the unenv cloudflare preset should configure bundler to alias node:process (to unenv) or use external one. If it is aliased to unenv, it will make a recursive situation and if it is external unenv is not effective.

if we resolve the process.env issue as suggested above, we might not need to import `node:process` at all, but if we do, we would only use it for APIs that are natively implemented in workerd, which will remain exposed once we implement the hybrid polyfills as described in https://github.com/unjs/unenv/issues/181

so I think the last remaining thing here is to resolve the two issues I broght up with the current process.env polyfill (1/ removal of globalThis fallback 2/ swapping the order in read operations).

src/runtime/polyfill/cloudflare.ts
pi0: nitro has not yet adopted `cloudflare` preset of unenv so it won't affect legacy users 🤞🏼

pi0: regarding issue seems valid (and honestly this polyfill is hacky anyway),  i guess as long as we have `__env__` it will override it right?

pi0: BTW there is almost same impl in [std-env](https://github.com/unjs/std-env/blob/main/src/env.ts) (standalone util). We can probably test different scenarios better there using vitest and backport to unenv. (I plan to merge at some point in the future)

IgorMinar: how about this: https://github.com/unjs/unenv/pull/186

jculvey:
Update: I've changed this PR so that the polyfill simply sets `globalThis.__env__` to empty object.

src/runtime/polyfill/cloudflare.ts
```diff
@@ -0,0 +1 @@
+(globalThis as any).__env__ = {};
```
pi0, line 1: Maybe add a comment what it does?

pi0, line 1:
```suggestion
// Make sure env polyfill won't fallback to legacy `globalThis` used for service-worker format
(globalThis as any).__env__ = {};
```

pi0 (approved):

PR #184, opened 2024-05-15
feat(node:process): prefer `queueMicrotask` for `process.nextTick` polyfill

pi0:
Idea from #183 use [`queueMicrotask`](https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask) for [approximate `process.nextTick()`](https://nodejs.org/api/process.html#when-to-use-queuemicrotask-vs-processnexttick) in Node.js when available.

### DIFF
--- a/src/runtime/node/process/_process.ts
+++ b/src/runtime/node/process/_process.ts
@@ -135,7 +135,7 @@ function drainQueue() {
 }
 
 process.nextTick = function (fun) {
-  // https://github.com/cloudflare/workerd/blob/main/src/node/internal/process.ts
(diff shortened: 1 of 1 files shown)


IgorMinar (approved):
this works for me, but keep in mind that it's a potential breaking change as the previous implementation was not compliant and some apps that use unenv might now depend on the non-compliant behavior. I don't know how much you care...

one more thing for you to consider is keeping the clean and old implementation as two alternative implementations, something along the lines:

```
process.nextTick = globalThis.queueMicrotask ?
    function nextTick(cb: Function, ...args: unknown[]) {
      globalThis.queueMicrotask(() => { cb(...args); });
    } : ....their existing code here ....
```

PR #185, opened 2024-05-15
feat(node:process): add additional polyfill coverage for process

jculvey:
### ❓ Type of change

- [x] 👌 Enhancement (improving an existing functionality like performance)

### 📚 Description

Adds additional polyfill coverage for `process`.

### 📝 Checklist

### DIFF
--- a/src/runtime/node/process/_process.ts
+++ b/src/runtime/node/process/_process.ts
@@ -5,8 +5,6 @@ import mock from "../../mock/proxy";
 import empty from "../../mock/empty";
 import { notImplemented } from "src/runtime/_internal/utils";
 
-export const process = {} as typeof globalThis.process;
-
 // Cached from whatever global is present so that test runners that stub it
 // don't break things.  But we need to wrap it in a try catch in case it is
 // wrapped in strict mode code which doesn't define any globals.  It's inside a
@@ -159,7 +157,7 @@ function _nextTick() {
   });
 }
 
-process.nextTick = globalThis.queueMicrotask ? _nextTick : _nextTickLegacy;
+const nextTick = globalThis.queueMicrotask ? _nextTick : _nextTickLegacy;
 
 // v8 likes predictible objects
 function Item(fun, array) {
@@ -169,14 +167,14 @@ function Item(fun, array) {
 Item.prototype.run = function () {
   this.fun.apply(null, this.array);
 };
-process.title = "unenv";
+const title = "unenv";
 
 const _envShim = Object.create(null);
 const _processEnv = globalThis.process?.env;
 const _getEnv = (useShim: boolean) =>
   _processEnv || globalThis.__env__ || (useShim ? _envShim : globalThis);
 
-process.env = new Proxy(_envShim, {
+const env = new Proxy(_envShim, {
   get(_, prop) {
     const env = _getEnv();
     return env[prop] ?? _envShim[prop];
@@ -200,50 +198,50 @@ process.env = new Proxy(_envShim, {
   },
 });
 
-process.argv = [];
+const argv = [];
 // @ts-ignore
-process.version = ""; // empty string to avoid regexp issues
+const version = ""; // empty string to avoid regexp issues
 // @ts-ignore
-process.versions = {};
+const versions = {};
 
 function noop() {
   return process;
 }
 
-process.on = noop;
-process.addListener = noop;
-process.once = noop;
-process.off = noop;
-process.removeListener = noop;
-process.removeAllListeners = noop;
+const on = noop;
+const addListener = noop;
+const once = noop;
+const off = noop;
+const removeListener = noop;
+const removeAllListeners = noop;
 // @ts-ignore
-process.emit = noop;
-process.prependListener = noop;
-process.prependOnceListener = noop;
+const emit = noop;
+const prependListener = noop;
+const prependOnceListener = noop;
 
-process.listeners = function (name) {
+const listeners = function (name) {
   return [];
 };
-process.listenerCount = () => process.listeners().length;
+const listenerCount = () => process.listeners().length;
 
 // @ts-ignore
-process.binding = function (name) {
+const binding = function (name) {
   throw new Error("[unenv] process.binding is not supported");
 };
 
-let cwd = "/";
-process.cwd = function () {
-  return cwd;
+let _cwd = "/";
+const cwd = function () {
+  return _cwd;
 };
-process.chdir = function (dir) {
-  cwd = dir;
+const chdir = function (dir) {
+  _cwd = dir;
 };
-process.umask = function () {
+const umask = function () {
   return 0;
 };
 
 // https://nodejs.org/api/process.html#processhrtime
-process.hrtime = function (startTime) {
+const hrtime = function (startTime) {
   const now = Date.now();
   // millis to seconds
   const seconds = Math.trunc(now / 1000);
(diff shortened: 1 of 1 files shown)


jculvey:
@pi0 please let me know if there's a better way to go about stubbing these or if there's anything here you like to see polyfilled differently.

src/runtime/node/process/_process.ts
```diff
@@ -262,3 +266,125 @@ process.hrtime.bigint = function () {
...
+process.setgid = function (val: number) {
+  gid = val;
+};
+
+let groups = [];
+process.getgroups = function () {
+  return groups;
+};
+process.setgroups = function (val: number[]) {
+  groups = val;
```
IgorMinar, line 299: since this is supported in node only if you have the right privileges I think we should just error with the unimplemented fn as you used elsewhere.

this is what node does unless you run it as root:

```
> process.setgroups([20])
Uncaught Error: EPERM, Operation not permitted
    at process.setgroups (node:internal/bootstrap/switches/does_own_process_state:77:20) {
  errno: 1,
  code: 'EPERM',
  syscall: 'setgroups'
}
```
The same goes for
- [ ] setegid
- [ ] seteuid
- [ ] setgid
- [ ] setuid

src/runtime/node/process/_process.ts
```diff
@@ -262,3 +266,125 @@ process.hrtime.bigint = function () {
...
+
+process.assert = notImplemented("process.assert");
+process.abort = notImplemented("process.abort");
+process.allowedNodeEnvironmentFlags = [];
+process.arch = "";
+process.argv0 = "";
+process.config = empty;
+process.connected = false;
+process.constrainedMemory = 0;
+process.cpuUsage = notImplemented("process.cpuUsage") as boolean;
```
IgorMinar, line 341: `as boolean` seems wrong here. this api seems to be a function that returns an object: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/bfdcd1c2bad2ed372ebf698103824f4171e5e972/types/node/process.d.ts#L1330

src/runtime/node/process/_process.ts
```diff
@@ -262,3 +266,125 @@ process.hrtime.bigint = function () {
...
+process.emitWarning = noop;
+process.eventNames = notImplemented("process.eventNames");
+process.execArgv = [];
+process.execPath = "";
+process.exit = notImplemented("process.exit");
+process.features = Object.create({
+  inspector: false,
+  debug: false,
+  uv: false,
+  ipv6: false,
```
IgorMinar, line 354: I don't know how much this matters / how much usage this gets, but google search doesn't reveal significant usage, so I think it's fine that this is always set to `false`

src/runtime/node/process/_process.ts
```diff
@@ -262,3 +266,125 @@ process.hrtime.bigint = function () {
...
+  tls_alpn: false,
+  tls_sni: false,
+  tls_ocsp: false,
+  tls: false,
+  cached_builtins: false,
+});
+process.getActiveResourcesInfo = [];
+process.getMaxListeners = notImplemented("process.getMaxListeners");
+process.kill = notImplemented("");
+process.listenerCount = () => process.listeners().length;
```
IgorMinar, line 364: consider moving this below line #224 to keep all the listener methods together

https://github.com/unjs/unenv/blob/8d8548427f3a10b12115c76928af7ef631dd0078/src/runtime/node/process/_process.ts#L224

src/runtime/node/process/_process.ts
```diff
@@ -262,3 +266,125 @@ process.hrtime.bigint = function () {
...
+  reportOnFatalError: false,
+  reportOnSignal: false,
+  reportOnUncaughtException: false,
+  signal: "",
+  writeReport: notImplemented("process.report.writeReport"),
+});
+process.resourceUsage = notImplemented("process.resourceUsage");
+process.setMaxListeners = notImplemented("process.setMaxListeners");
+process.setSourceMapsEnabled = notImplemented("process.setSourceMapsEnabled");
+process.title = "";
```
IgorMinar, line 386: title is already defined earlier in this file...

IgorMinar (approved):
I have a few small suggestions, but overall this looks good to me.

Let's have @pi0 take a look as well before we merge this.

src/runtime/node/process/_process.ts
```diff
@@ -262,3 +266,125 @@ process.hrtime.bigint = function () {
...
+let euid = 0;
+process.geteuid = function () {
+  return euid;
+};
+process.seteuid = function (val: number) {
+  euid = val;
+};
+
+let gid = 0;
+process.getgid = function () {
```
pi0, line 287: We could default to a non zero value (like `1000`) for instead of `0` (root). In case of any dependency of this to do an operation I think it might be safer..

src/runtime/node/process/_process.ts
pi0: throwing not implemented for `set` methods makes sense (I have to span a windows machine but I guess that's Node.js behavior too when called on windows/android)

pi0: Testing on stackblitz, they also throw `EPERM` for `setgroups`. For `get` methods, they return a non zero number which looks safe to me if sensable default to try making codes work as much as possible.

src/runtime/node/process/_process.ts
pi0: (we could also use null/undefiend too btw to clearly indicate these values have no meaningful value)

src/runtime/node/process/_process.ts
```diff
@@ -262,3 +266,125 @@ process.hrtime.bigint = function () {
...
+  tls: false,
+  cached_builtins: false,
+});
+process.getActiveResourcesInfo = [];
+process.getMaxListeners = notImplemented("process.getMaxListeners");
+process.kill = notImplemented("");
+process.listenerCount = () => process.listeners().length;
+process.memoryUsage = notImplemented("process.memoryUsage");
+process.openStdin = notImplemented("process.openStdin");
+process.pid = 0;
```
pi0, line 367: Would be nicer to use non zero values (pid zero is for init process)

pi0:
Thanks dear @jculvey for nice work on it.

One generic suggestion, if possible, can you try to refactor process assignments into one object assignment? v8 engine does not likes object reshaping (adding props) also it might help to prevent typos like https://github.com/unjs/unenv/pull/185#discussion_r1602241525

src/runtime/node/process/_process.ts
```diff
@@ -262,3 +266,125 @@ process.hrtime.bigint = function () {
...
+process.platform = "";
+process.ppid = 0;
+process.rawListeners = notImplemented("process.rawListeners");
+process.release = empty;
```
pi0, line 371: It would be nice to make source at least `name` has empty value in case of string operations

src/runtime/node/process/_process.ts
```diff
@@ -258,7 +261,197 @@ process.hrtime = function (startTime) {
...
+  ipv6: undefined,
+  tls_alpn: undefined,
+  tls_sni: undefined,
+  tls_ocsp: undefined,
+  tls: undefined,
+  cached_builtins: undefined,
+});
+const getActiveResourcesInfo = [];
+const getMaxListeners = notImplemented("process.getMaxListeners");
+const kill = notImplemented("");
```
pi0, line 337:
```suggestion
const kill = notImplemented("process.kill");
```

src/runtime/node/process/_process.ts
```diff
@@ -258,7 +261,197 @@ process.hrtime = function (startTime) {
...
+const resourceUsage = notImplemented("process.resourceUsage");
+const setegid = notImplemented("process.setegid");
+const seteuid = notImplemented("process.seteuid");
+const setgid = notImplemented("process.setgid");
+const setgroups = notImplemented("process.setgroups");
+const setuid = notImplemented("process.setuid");
+const setMaxListeners = notImplemented("process.setMaxListeners");
+const setSourceMapsEnabled = notImplemented("process.setSourceMapsEnabled");
+const stdout = mock.__createMock__("process.stdout");
+const stderr = mock.__createMock__("process.stdout");
```
pi0, line 370:
```suggestion
const stderr = mock.__createMock__("process.err");
```

src/runtime/node/process/_process.ts
```diff
@@ -258,7 +261,197 @@ process.hrtime = function (startTime) {
...
+const stdin = mock.__createMock__("process.stdout");
```
pi0, line 371:
```suggestion
const stdin = mock.__createMock__("process.stdin");
```

src/runtime/node/process/_process.ts
```diff
@@ -258,7 +261,197 @@ process.hrtime = function (startTime) {
...
+const traceDeprecation = false;
+const uptime = 1000;
```
pi0, line 373:
```suggestion
const uptime = 0;
```

pi0 (approved):
Thanks!

PR #186, opened 2024-05-15
fix(node): rewrite the process.env polyfill

IgorMinar:
### 🔗 Linked issue

https://github.com/unjs/unenv/pull/183

### ❓ Type of change

### 📚 Description

The previous polyfill was unnecessarily complicated and hard to understand. Additionally it also used fallback on globalThis to read and write env variables, which in the past was used to read env variables in Cloudflare Workers in the case where workers were using the legacy service worker format. This style of workers is not actively used any more so this fallback is not needed, and in fact is undesirable as we never want to accidentaly write this globalThis.foo as a result of an assignment to process.env.foo.

The nitro utilized globalThis.__env__ fallback is preserved as to no break nitro users.

### 📝 Checklist

- [x] I have linked an issue or discussion.

### DIFF
--- a/src/runtime/node/process/_process.ts
+++ b/src/runtime/node/process/_process.ts
@@ -167,34 +167,14 @@ Item.prototype.run = function () {
 };
 process.title = "unenv";
 
-const _envShim = Object.create(null);
-const _processEnv = globalThis.process?.env;
-const _getEnv = (useShim: boolean) =>
-  _processEnv || globalThis.__env__ || (useShim ? _envShim : globalThis);
+process.env = 
+  // preserve process.env if it exists
+  globalThis.process?.env ||
+  // use global __env__ if exists, this is used by nitro to inject env vars from build
+  globalThis.__env__ ||
+  // fall back on an empty object
+  Object.create(null);
 
-process.env = new Proxy(_envShim, {
-  get(_, prop) {
-    const env = _getEnv();
-    return env[prop] ?? _envShim[prop];
-  },
-  has(_, prop) {
-    const env = _getEnv();
-    return prop in env || prop in _envShim;
-  },
-  set(_, prop, value) {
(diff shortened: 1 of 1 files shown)


pi0:
I appreciate your attempt to improve this but it probably won't work as expected anymore.

For cloudflare presets, we assign to `globalThis.__env__` after first `fetch` event. It is after when the polyfill being initialized and the ref to `globalThis.__env__` (undefined) at time of startup won't be updated to new `process.env`.

Besides, since the polyfill needs to cover other runtimes such as [`Deno.env`](https://docs.deno.com/runtime/manual/basics/env_variables) which has explicit getters and setters proxy will be needed for them.

(And to add to the complexity, there are frameworks such as astro use [`import.meta.env.*`](https://github.com/withastro/roadmap/issues/853) so we cannot really simplify env support that much)

I think simplest way to disable legacy cloudflare-service-worker behavior that picks the `globalThis.*` would be to assign an empty object to `globalThis.__env__` to disable current behavior.

---

We can introduce a trimmed-down polyfill via `polyfill/cloudflare.ts` being introduced in #183 if you like but still we need a getter method at least to allow injection after fetch.

IgorMinar:
Interesting. Thanks for the explanation @pi0. Let me think about this a bit more.

IgorMinar:
I just noticed this line of code: https://github.com/unjs/unenv/blob/e837dfab520d0e31a259e7868509ac40478a614e/src/runtime/node/process/_env.ts#L2

which when combined with `process.getBuiltinModule` from https://github.com/cloudflare/workerd/pull/2147 should make it unnecessary for us to make any changes with unenv.

I'm going to close this PR for now and we can revisit it later.

I noticed that https://github.com/unjs/unenv/pull/183  was merged, which I'm not very keen on, and I'll start a convo on discord about considering to revert it.

PR #189, opened 2024-05-20
feat(node:console): polyfill `node:console` module

jculvey:
### ❓ Type of change

### 📚 Description

Adds a polyfill for the `node:console` module.

### 📝 Checklist


src/runtime/node/console/index.ts
```diff
@@ -0,0 +1,230 @@
+import type console from "node:console";
+import { createNotImplementedError } from "../../_internal/utils";
+import mock from "../../mock/proxy";
+
+export const assert: typeof console.assert = (
+  ...args: Parameters<(typeof globalThis)["console"]["assert"]>
+) => {
+  if (globalThis.console && globalThis.console.assert) {
+    return globalThis.console.assert(...args);
+  }
```
IgorMinar, line 10: you could simplify this a bit with safe navigation operator:

```suggestion
  if (globalThis.console?.assert) {
    return globalThis.console.assert(...args);
  }
```

and use this pattern throughout this whole file

src/runtime/node/console/index.ts
```diff
@@ -0,0 +1,230 @@
...
+  if (globalThis.console && globalThis.console.warn) {
+    return globalThis.console.warn(...args);
+  }
+  createNotImplementedError("console.warn");
+};
+
+export const Console: typeof console.Console =
+  globalThis.console && globalThis.console.Console
+    ? globalThis.console.Console
+    : mock.__createMock__("console.Console");
```
IgorMinar, line 204: You can use nullish coalescing operator here to simplify the code:

```suggestion
export const Console: typeof console.Console =
    // Node.js exposes the Console constructor on the console instance 🤷🏼‍♂️
    // https://nodejs.org/api/console.html#class-console
    globalThis.console?.Console ?? mock.__createMock__("console.Console");
```

I'm however not convinced that we should try to grab the constructor from globalThis.console at all as that's a node.js-only thing and is not spec compliant, so this would only work in node.js environment and `nodeless` which this polyfill is trying to backfill. So it's most likely never going to be used, which makes me think we should just always return a mock. But I wonder what does @pi0 think about this one.

IgorMinar (approved):
I only found a few small issues and change requests. Overall I really like this approach. Thanks @jculvey!

IgorMinar:
LGTM! I'm guessing that we'll need to squash this PR before/during merging. But I'm not quite sure what's the norm in this repo.

src/runtime/node/console/index.ts
```diff
@@ -0,0 +1,228 @@
+import type console from "node:console";
+import { createNotImplementedError } from "../../_internal/utils";
+import mock from "../../mock/proxy";
+
+export const assert: typeof console.assert = (
```
pi0, line 5: Fallback could be like `const assert = globalThis.console?.assert || notImplemented()`

src/runtime/node/console/index.ts
pi0: That's true indeed it is not specific but also like this because allows to optionally polyfill by summing in global `console` as an escape hatch.

pi0:
>  I'm guessing that we'll need to squash this PR before/during mergin

Across unjs repos, squash is default :)

src/runtime/node/console/index.ts
IgorMinar: I like this as it simplifies the code further. @jculvey can you please update the pr following this pattern?

I think it would look something like:

```
export const assert: typeof console.assert = globalThis.console?.assert ?? notImplemented("console.assert");
```

src/runtime/node/console/index.ts
jculvey: Sounds good, I've changed it back to the original implementation that tries to use `global.console.Console` and falls back to a mock. LMK if this works for y'all

pi0:
Few suggestions:

- We can cache `const _console = globalThis.console` once in top to reduce bundle impact little bit
- For various console log methods such as `info`, `debug`, .. we can redirect them to `console.log` in minimum runtimes rather than throwing error
- I would use explicit no-op instead of not-implemented throws for informational things such as `groupStart`. They won't be really needed but if a library uses them, it breaks runtime unnecessarily and they are only for formatting

src/runtime/node/console/index.ts
```diff
@@ -0,0 +1,60 @@
+import type console from "node:console";
+import mock from "../../mock/proxy";
+
+const _console = globalThis.console;
+
+const noop = () => {};
```
pi0, line 6: (we have `mock/noop` saves few bytes)

src/runtime/node/console/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const log: typeof console.log = _console?.log ?? noop;
+const info: typeof console.info = _console?.info ?? log;
+const warn: typeof console.warn = _console?.warn ?? log;
+const debug: typeof console.debug = _console?.debug ?? log;
+const error: typeof console.error = _console?.error ?? log;
+
+const assert: typeof console.assert = _console?.assert ?? noop;
```
pi0, line 14: i think asset could be marked as unimplemented

src/runtime/node/console/index.ts
```diff
@@ -0,0 +1,60 @@
...
+const countReset: typeof console.countReset = _console?.countReset ?? noop;
+const dir: typeof console.dir = _console?.dir ?? noop;
+const dirxml: typeof console.dirxml = _console?.dirxml ?? noop;
+const group: typeof console.group = _console?.group ?? noop;
+const groupEnd: typeof console.groupEnd = _console?.groupEnd ?? noop;
+const groupCollapsed: typeof console.groupCollapsed =
+  _console?.groupCollapsed ?? noop;
+const profile: typeof console.profile = _console?.profile ?? noop;
+const profileEnd: typeof console.profileEnd = _console?.profileEnd ?? noop;
+const table: typeof console.table = _console?.table ?? noop;
```
pi0, line 26: this could be safely fallback to `log` table would be array of arrays instead of pretty output

src/runtime/node/console/index.ts
```diff
@@ -0,0 +1,60 @@
...
+const time: typeof console.time = _console?.time ?? noop;
+const timeEnd: typeof console.timeEnd = _console?.timeEnd ?? noop;
+const timeLog: typeof console.timeLog = _console?.timeLog ?? noop;
+const timeStamp: typeof console.timeStamp = _console?.timeStamp ?? noop;
+const trace: typeof console.trace = _console?.trace ?? noop;
```
pi0, line 31: trace is log + stacktrack. we can easily polyfill using `newError` or simply fallback to `log`

pi0 (approved):

pi0:
Thanks!

PR #190, opened 2024-05-20
feat(node:os): polyfill `node:os` module

jculvey:
### ❓ Type of change

- [x] 👌 Enhancement (improving an existing functionality like performance)

### 📚 Description

Adds a polyfill for the `node:os` module.

Curious for feedback on `EOL` and `devNull`. I don't see an "unimplemented" pattern in place for handling string constants exported by modules other than specifying a dummy value. LMK if y'all think there's a better value to use for either of these.

### 📝 Checklist


src/runtime/node/os/index.ts
```diff
@@ -0,0 +1,60 @@
...
+export const tmpdir: typeof os.tmpdir = notImplemented("os.tmpdir");
+export const totalmem: typeof os.totalmem = notImplemented("os.totalmem");
+export const type: typeof os.type = notImplemented("os.type");
+export const uptime: typeof os.uptime = notImplemented("os.uptime");
+export const userInfo: typeof os.userInfo = notImplemented("os.userInfo");
+export const version: typeof os.version = notImplemented("os.version");
+
+export const constants: typeof os.constants =
+  mock.__createMock__("os.constants");
+export const devNull: typeof os.devNull = "";
```
IgorMinar, line 33: I think this should be a real value. `"/dev/null"` doesn't seem too crazy/problematic.

src/runtime/node/os/index.ts
```diff
@@ -0,0 +1,60 @@
...
+export const EOL: typeof os.EOL = "\n";
```
IgorMinar, line 34: lgtm

IgorMinar (approved):
LGTM except for `devNull`.

It would be good to expand the commit message to expand on why we are replacing the default auto-mock with this more explicit mock.

The main usecase is to support destructured ESM imports:
```
import { devNull } from 'os';
```

Which currently fail because the auto-mock can't be statically analyzed:
Screenshot 2024-05-20 at 12 53 26 PM

src/runtime/node/os/index.ts
```diff
@@ -0,0 +1,60 @@
+import type os from "node:os";
+import { notImplemented } from "../../_internal/utils";
+import mock from "../../mock/proxy";
+
+export const arch: typeof os.arch = notImplemented("os.arch");
+export const availableParallelism: typeof os.availableParallelism =
+  notImplemented("os.availableParallelism");
+export const cpus: typeof os.cpus = notImplemented("os.cpus");
```
pi0, line 8: Some simple items can be simply non-throwing and return a  fallback value IMO. It increases compatibility (end-users don't have a choice when they need a lib that relyins on these APIs. If it throws it does not fixes anything but fallback value would)

IgorMinar: hmmm... do you have a suggested value here?

When I test this, the output on my laptop is not trivial:

```
> require('os').cpus()
[
  {
    model: 'Apple M1 Pro',
    speed: 2400,
    times: {
      user: 148364570,
      nice: 0,
      sys: 97148930,
      idle: 304228910,
      irq: 0
    }
  },
  {
    model: 'Apple M1 Pro',
    speed: 2400,
    times: {
      user: 147962800,
      nice: 0,
      sys: 95743880,
      idle: 309692400,
      irq: 0
    }
  },
  {
    model: 'Apple M1 Pro',
    speed: 2400,
    times: {
      user: 134111910,
      nice: 0,
      sys: 35374800,
      idle: 402528560,
      irq: 0
    }
  },
  {
    model: 'Apple M1 Pro',
    speed: 2400,
    times: { user: 84991730, nice: 0, sys: 25427380, idle: 463970630, irq: 0 }
  },
  {
    model: 'Apple M1 Pro',
    speed: 2400,
    times: { user: 55427660, nice: 0, sys: 14547330, idle: 506192030, irq: 0 }
  },
  {
    model: 'Apple M1 Pro',
    speed: 2400,
    times: { user: 39030960, nice: 0, sys: 10133530, idle: 527638760, irq: 0 }
  },
  {
    model: 'Apple M1 Pro',
    speed: 2400,
    times: { user: 40421640, nice: 0, sys: 6451750, idle: 530715920, irq: 0 }
  },
  {
    model: 'Apple M1 Pro',
    speed: 2400,
    times: { user: 19716300, nice: 0, sys: 3817220, idle: 554515890, irq: 0 }
  },
  {
    model: 'Apple M1 Pro',
    speed: 2400,
    times: { user: 12116580, nice: 0, sys: 2508300, idle: 563667690, irq: 0 }
  },
  {
    model: 'Apple M1 Pro',
    speed: 2400,
    times: { user: 9789590, nice: 0, sys: 2036340, idle: 566538860, irq: 0 }
  }
]
```

Since this change is not making the situation worse than it is now with the current module level mock, I suggest we merge this as is to increase the "presence" coverage, and then do a second pass and replace these throwing functions with some mock value.

In general, our strategy is to get to 100% of API "presence" coverage quickly, and then do a second pass of filling in the implementation where it's simple or important for real apps/libraries.

pi0: please note that currently all missing exports _are_ polyfilled (via proxy) but won't throw an error. It can potentially (and in case of this module i suspect more likely) that we break current projects because the polyfill changes from noop to error throws.

[stackblitz for example](https://stackblitz.com/edit/stackblitz-starters-ugx7ss?description=Starter%20project%20for%20Node.js,%20a%20JavaScript%20runtime%20built%20on%20Chrome%27s%20V8%20JavaScript%20engine&file=index.js,package.json&title=node.new%20Starter) returns random 8 cores.

pi0:
@jculvey @IgorMinar i've added implementation too. Constants is little bit long (but treehsakable and when minified small). Wdyt?

IgorMinar (approved):
LGTM

IgorMinar:
@jculvey if you are happy with this PR please go ahead and merge

jculvey:
> @jculvey if you are happy with this PR please go ahead and merge

Yup LGTM. Thanks again @pi0 for fixing this up!

PR #191, opened 2024-05-20
feat(cloudflare): alias `node:assert/strict` to `node:assert`

jculvey:
### ❓ Type of change

### 📚 Description

Since the `workerd` implementation of the `node:assert` module already has strict semantics, we simply need to alias `node:assert/strict` to `node:assert` to get coverage for the module.

### 📝 Checklist


IgorMinar (commented):
According to https://nodejs.org/api/assert.html#strict-assertion-mode 

> In strict assertion mode, non-strict methods behave like their corresponding strict methods. For example, [assert.deepEqual()](https://nodejs.org/api/assert.html#assertdeepequalactual-expected-message) will behave like [assert.deepStrictEqual()](https://nodejs.org/api/assert.html#assertdeepstrictequalactual-expected-message).

So implementing the real thing is actually less code than implementing the mocks. Could you please reexport the `node:assert` symbols here instead?

jculvey:
> According to https://nodejs.org/api/assert.html#strict-assertion-mode
> 
> > In strict assertion mode, non-strict methods behave like their corresponding strict methods. For example, [assert.deepEqual()](https://nodejs.org/api/assert.html#assertdeepequalactual-expected-message) will behave like [assert.deepStrictEqual()](https://nodejs.org/api/assert.html#assertdeepstrictequalactual-expected-message).
> 
> So implementing the real thing is actually less code than implementing the mocks. Could you please reexport the `node:assert` symbols here instead?

Yeah sure thing. Should I do that a Cloudflare specific polyfill (since `nodeless` doesn't have `node:assert`), or should we use the try/fallback approach and use these if  detect `assert` and fallback to `notImplemented`?

pi0 (approved):

PR #193, opened 2024-05-21
fix(nodelss): inject and polyfill node `global`

IgorMinar:
Since `global` comes mainly from Node.js, we should by default provide it's polyfill in the nodeless preset so that other presets don't need to define it.

For example Deno preset shouldn't need to define that it needs the global polyfill, as we expect the target environment to compose `env(nodeless, deno)` together.

### 🔗 Linked issue

n/a

### ❓ Type of change

### 📝 Checklist


src/presets/nodeless.ts
```diff
@@ -49,11 +49,12 @@ const nodeless: Preset & { alias: Map<string, string> } = {
   },
 
   inject: {
+    global: "unenv/runtime/polyfill/global",
```
pi0, line 52: (please try if you can) i think this can be `global-this` one to avoid side-effect:

```suggestion
    global: "unenv/runtime/polyfill/global-this",
```

IgorMinar: Do you mean this defineProperty call?

https://github.com/IgorMinar/unenv/blob/2f980cb4c9425c4a8fdabfb10be30587b1e58fc3/src/runtime/polyfill/global.ts#L5

That call does seem wrong, as the `global` property of `globalThis` in Node 22 is enumerable:

```
Object.getOwnPropertyDescriptor(globalThis, 'global').enumerable
> true
```

I'm also not keen on exposing `window` and `self` as a side-effect of this polyfill, which don't belong to Node.

I do however want the side-effect of this polyfill to be that `globalThis` ends up having an enumerable `global` property just like it's in node.

I assume that it's not safe to change polyfill/global.ts in a breaking way. should I create a new file polyfill/node-global.ts?

pi0: What i mainly meant in this comment is that the `inject` option should not have side-effects. Current change replaces usage of `global` to `globalThis` + adds side-effect of polyfill (whatever it is) on usage which would be extra.

---

On improving polyfill i generally agree that `global` can be polyfilled to `globalThis` now that browser support is wide enough for `globalThis`. I think we can introduce it in another (but breaking) change to drop polyfill `[global].window`.

IgorMinar: I see... is this so that the inject is tree-shakeable or do you in general want injects to be side-effect free so that the affect only the code in the bundle and not code outside of the bundle?

I'm asking because I've just implemented tree-shakeable injectes with side-effects in wrangler: https://github.com/cloudflare/workers-sdk/pull/5878/files#diff-d5f05e17ffb161174d2bfea37bf23d39679290784e9aff28ac77bc06e3492459R84-R104

The main reason for this is that I do want the global side-effects to be present, but only if we detect usage of the global. This way we don't pay the cost of polyfills if they are not needed, and for rare cases when a 3rd party code accesses globals via patterns that are hard to statically analyze, users can inject force the polyfill side-effects to be present by adding user-land code that triggers the injection.

pi0: Sometimes bundlers make strange behavior (especially with chunking) when there is a module with side-effects in the dependency path.

I see the idea of on-demand polyfill but it would introduce an implicit behavior. Consider a scenario in which users use two libraries one is statically analyzable and because of lib 1 and injects, we (globally) polyfill something that fixes usage in lib 2 that cannot be statically analyzable by the inject plugin. now if libraries order changes, lib 1 is not used, or even simply the chunk loading order changes polyfill will be broken. 

For global polyfill really doesn't matter for me that now we are both polyfill + inject but for things like `Buffer` (in other runtimes) it is costly enough that we _only_ rely on inject for the on-demand cost savings.

IgorMinar: So your argument is two fold:

1. size-savings since bundlers would not tree-shake code if they see the side effect
    - that's valid but can be worked around using the @__PURE__ annotation that is now supported by all bundlers and has been battle tested by Angular for 5+ years.
2. accidentally breaking 2nd app in the came vm if it depends but doesn't declare the dependency on the global side-effect
    - again, this is a valid concern, but the situation is not any different if the first app uses `polyfills` instead of `injects` which makes the 2nd app work.

That being said, I think it's fine to refactor the code to make `injects` side-effect-free, and `polyfills` to be side-effecty. This distinction should really be documented though as it's not obvious at all and it took us a good few days to figure it out, and then I still got it wrong in this PR.

IgorMinar: how about this approach: https://github.com/unjs/unenv/pull/193/commits/ac20651f14893e9cf6e6a77ae42a6bfccd8558d5

IgorMinar: > I don't mean polyfill by inject but polyfill in entrypoint (nitro always adds polyfills to the entrypoints explicitly to make sure of ordering and probably we need to generalize this to make sure stable support. see [here](https://github.com/unjs/nitro/blob/main/src/presets/cloudflare/runtime/cloudflare-pages.ts#L1) and [here](https://github.com/unjs/nitro/blob/6c3e0805d38c80fcad9430d1e5beeeee65bbbaa5/src/rollup/config.ts#L319))

As you know, one big downside of this approach is that you'll end up with all polyfills even if you don't need them, which I'd like to avoid in Wrangler.

pi0: > Anyway, how about this approach: https://github.com/unjs/unenv/commit/ac20651f14893e9cf6e6a77ae42a6bfccd8558d5 

love it!

src/runtime/polyfill/node-global.ts
```diff
@@ -0,0 +1,2 @@
+// publish "global" on globalThis
+globalThis.global = globalThis;
```
pi0, line 2:
```suggestion
if (!('global' in globalThis)) {
  globalThis.global = globalThis;
}
```

src/runtime/node/global.ts
pi0, line 1: one last thing, since `runtime/node/*` is mapped to `node:` imports, maybe move it to `polyfill/global-this-native` for more clarity?

src/runtime/polyfill/node-global.ts
pi0, line 1: typo: node-global

src/runtime/node/global.ts
IgorMinar: I find that confusing. I though `polyfills` directory was for side-effecty code :)

I could move it to `runtime/node/globals/global.ts` - would that help?

pi0: Agree. We have the side-effect-less `polyfill/global-this` there too btw which was confusing. Maybe `runtime/globals/global-this` + `runtime/globals/global-this-native` (i use `native` to indicate we pick `globalThis` whatever is defined in runtime without search)

IgorMinar: Shouldn't the filename reflect *what* is being polyfilled, rather than *how* it's being polyfilled?

pi0: we have `polyfill/global-this` which is also side-effect free but let's iterate later 👍🏼

pi0 (approved):

PR #194, opened 2024-05-21
refactor: pass unimplemented types as generic

pi0:
fixes current type issues with function overloads in node defenitions


jculvey:
I'm guessing this will fix the CI failures that started popping up like [this one](https://github.com/unjs/unenv/actions/runs/9180056140/job/25243667828?pr=189), right?

PR #197, opened 2024-05-22
feat(node:cluster): polyfill `node:cluster` module

jculvey:
Replaces the current auto-mocking of 'cluster' to support destructured ESM imports and allow for functional polyfill coverage in the future.

### ❓ Type of change

### 📚 Description

Adds a polyfill for the `node:cluster` module.

### 📝 Checklist

### DIFF
--- a/src/runtime/node/cluster/index.ts
+++ b/src/runtime/node/cluster/index.ts
@@ -22,8 +22,8 @@ const _cluster = new EventEmitter() as MutableCluster;
 export const disconnect: typeof cluster.disconnect = noop;
 export const fork: typeof cluster.fork = () =>
   mock.__createMock__("cluster.Worker");
-export const isPrimary: typeof cluster.isPrimary = true;
 export const isMaster: typeof cluster.isMaster = true;
+export const isPrimary: typeof cluster.isPrimary = true;
 export const isWorker: typeof cluster.isWorker = false;
 export const SCHED_NONE: typeof cluster.SCHED_NONE = 1;
 export const SCHED_RR: typeof cluster.SCHED_RR = 2;
@@ -36,8 +36,8 @@ export const Worker: typeof _Worker = mock.__createMock__("cluster.Worker");
 
 _cluster.disconnect = disconnect;
 _cluster.fork = fork;
-_cluster.isPrimary = isPrimary;
 _cluster.isMaster = isMaster;
+_cluster.isPrimary = isPrimary;
 _cluster.isWorker = isWorker;
 _cluster.SCHED_NONE = SCHED_NONE;
 _cluster.SCHED_RR = SCHED_RR;
--- a/src/runtime/node/cluster/index.ts
+++ b/src/runtime/node/cluster/index.ts
@@ -1,43 +1,27 @@
-import { notImplemented } from "src/runtime/_internal/utils";
 import noop from "../../mock/noop";
 import mock from "../../mock/proxy";
 import type cluster from "node:cluster";
 import type { Cluster, Worker as _Worker } from "node:cluster";
 import { EventEmitter } from "../events";
 
-const _cluster = new EventEmitter() as Cluster;
-const _noop = () => {
-  return _cluster;
+/**
+ * Inspired by the original Node.js code in:
+ * https://github.com/nodejs/node/blob/main/lib/internal/cluster/primary.js
+ */
+
+// A mapped type used internally to allow assigning to readonly fields like `isPrimary`
+type MutableCluster = {
+  -readonly [key in keyof Cluster]: Cluster[key];
+} & {
+  Worker: typeof _Worker;
 };
 
-const on: typeof cluster.on = _noop;
-const addListener: typeof cluster.addListener = _noop;
-const once: typeof cluster.once = _noop;
-const removeListener: typeof cluster.removeListener = _noop;
-const removeAllListeners: typeof cluster.removeAllListeners = _noop;
-const emit: typeof cluster.emit = () => false;
-const off: typeof cluster.off = _noop;
-const prependListener: typeof cluster.prependListener = _noop;
-const prependOnceListener: typeof cluster.prependOnceListener = _noop;
-const listeners: typeof cluster.listeners = function (name) {
-  return [];
-};
-const listenerCount: typeof cluster.listenerCount = () => 0;
-const setMaxListeners: typeof cluster.setMaxListeners = notImplemented(
-  "cluster.setMaxListeners",
-);
-const getMaxListeners: typeof cluster.getMaxListeners = notImplemented(
-  "cluster.getMaxListeners",
-);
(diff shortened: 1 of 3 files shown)


src/runtime/node/cluster/index.ts
```diff
@@ -0,0 +1,82 @@
...
+import type cluster from "node:cluster";
+import type { Cluster, Worker as _Worker } from "node:cluster";
+import { EventEmitter } from "../events";
+
+const _cluster = new EventEmitter() as Cluster;
+const _noop = () => {
+  return _cluster;
+};
+
+const on: typeof cluster.on = _noop;
```
jculvey, line 13: I tried to copy the existing pattern from `process`, but there isn't a global `cluster` object.

src/runtime/node/cluster/index.ts
```diff
@@ -0,0 +1,82 @@
...
+const removeAllListeners: typeof cluster.removeAllListeners = _noop;
+const emit: typeof cluster.emit = () => false;
+const off: typeof cluster.off = _noop;
+const prependListener: typeof cluster.prependListener = _noop;
+const prependOnceListener: typeof cluster.prependOnceListener = _noop;
+const listeners: typeof cluster.listeners = function (name) {
+  return [];
+};
+const listenerCount: typeof cluster.listenerCount = () => 0;
+const setMaxListeners: typeof cluster.setMaxListeners = notImplemented(
```
IgorMinar, line 26: why are certain things noops and others throw? wouldn't throwing be considered a breaking change compared to the current auto-mocking behavior?

pi0:
Nice work. 

Checking [Node.js implementation](https://github.com/nodejs/node/blob/main/lib/internal/cluster/primary.js) it seems they directly assign certain properties to the instance of `EventEmitter`. perhaps we can make it closer to this?

src/runtime/node/cluster/index.ts
```diff
@@ -0,0 +1,82 @@
+import { notImplemented } from "src/runtime/_internal/utils";
```
IgorMinar, line 1:
```suggestion
/**
 * Inspired by the original Node.js code in: 
 * https://github.com/nodejs/node/blob/main/lib/internal/cluster/primary.js
 */

import { notImplemented } from "src/runtime/_internal/utils";
```

src/runtime/node/cluster/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  mock.__createMock__("cluster.Worker");
+export const isPrimary: typeof cluster.isPrimary = true;
+export const isMaster: typeof cluster.isMaster = true;
+export const isWorker: typeof cluster.isWorker = false;
+export const SCHED_NONE: typeof cluster.SCHED_NONE = 1;
+export const SCHED_RR: typeof cluster.SCHED_RR = 2;
+export const schedulingPolicy: typeof cluster.schedulingPolicy = SCHED_RR;
+export const settings: typeof cluster.settings = {};
+export const setupPrimary: typeof cluster.setupPrimary = noop;
+export const setupMaster: typeof cluster.setupMaster = noop;
```
IgorMinar, line 33:
```suggestion
export const setupMaster: typeof cluster.setupMaster = noop;
export const setupPrimary: typeof cluster.setupPrimary = noop;
```

src/runtime/node/cluster/index.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+// eslint-disable-next-line unicorn/prefer-event-target
+const _cluster = new EventEmitter() as MutableCluster;
+
+export const disconnect: typeof cluster.disconnect = noop;
+export const fork: typeof cluster.fork = () =>
+  mock.__createMock__("cluster.Worker");
+export const isPrimary: typeof cluster.isPrimary = true;
+export const isMaster: typeof cluster.isMaster = true;
```
IgorMinar, line 26:
```suggestion
export const isMaster: typeof cluster.isMaster = true;
export const isPrimary: typeof cluster.isPrimary = true;
```

src/runtime/node/cluster/index.ts
```diff
@@ -0,0 +1,51 @@
...
+// A mapped type used internally to allow assigning to readonly fields like `isPrimary`
+type MutableCluster = {
+  -readonly [key in keyof Cluster]: Cluster[key];
+} & {
+  Worker: typeof _Worker;
+};
+
+// eslint-disable-next-line unicorn/prefer-event-target
+const _cluster = new EventEmitter() as MutableCluster;
+
```
IgorMinar, line 21: did you intentionally skip:

_events
_eventsCount
_maxListeners

I think we should add them as they are part of the public api even though they are `_`-prefixed.

IgorMinar (commented):
Almost there!!

src/runtime/node/cluster/index.ts
jculvey: We get `_events` and `_maxListeners` for free now that we're extending `EventEmitter` (defined [here](https://github.com/unjs/unenv/blob/main/src/runtime/node/events/_events.ts#L37-L40)). They aren't defined on the `Cluster` type though. I can probably add `_eventsCount` there as well if that's cool with @pi0.

IgorMinar: I see! but don't you still need add `export const _events`, etc. to support named imports?

jculvey: Yeah good point...

jculvey: ok fixed

IgorMinar (approved):
LGTM. thanks you for the updates!

src/runtime/node/cluster/index.ts
```diff
@@ -0,0 +1,52 @@
...
+export const settings: typeof cluster.settings = {};
+export const setupPrimary: typeof cluster.setupPrimary = noop;
+export const setupMaster: typeof cluster.setupMaster = noop;
+export const workers: typeof cluster.workers = {};
+export const Worker: typeof _Worker = mock.__createMock__("cluster.Worker");
+
+// These 3 _functions don't exist on the EventEmitter type
+export const _events = (_cluster as any)._events;
+export const _eventsCount = (_cluster as any)._eventsCount;
+export const _maxListeners = (_cluster as any)._maxListeners;
```
pi0, line 36: I think only named export of this from `node:cluster` is the `_cluster` instance do we need additional exports?

IgorMinar: according to our node API surface scans all of these are needed: https://workers-nodejs-compat-matrix.pages.dev/

pi0: Strange but fair enough I can confirm with local Node.js (ESM) as well they get converted to named exports. Technically they are keys of default export...

pi0 (approved):

pi0:
Thanks.

PR #200, opened 2024-05-23
feat: polyfill `node:diagnostics_channel` module

jculvey:
### ❓ Type of change

### 📚 Description

Adds a polyfill for the `node:diagnostics_channel` module.

### 📝 Checklist

### DIFF
--- a/src/runtime/node/diagnostics_channel/channel.ts
+++ b/src/runtime/node/diagnostics_channel/channel.ts
@@ -1,3 +1,4 @@
+import { createNotImplementedError } from "src/runtime/_internal/utils";
 import noop from "../../mock/noop";
 import type diagnostics_channel from "node:diagnostics_channel";
 
@@ -41,7 +42,13 @@ export class Channel<StoreType, ContextType>
     }
   }
 
-  bindStore() {}
-  unbindStore() {}
-  runStores() {}
+  bindStore() {
+    createNotImplementedError("Channel.bindStore");
+  }
+  unbindStore() {
+    createNotImplementedError("Channel.unbindStore");
+  }
+  runStores() {
+    createNotImplementedError("Channel.runStores");
+  }
 }
--- a/src/runtime/node/diagnostics_channel/channel.ts
+++ b/src/runtime/node/diagnostics_channel/channel.ts
@@ -1,5 +1,4 @@
 import { createNotImplementedError } from "src/runtime/_internal/utils";
-import noop from "../../mock/noop";
 import type diagnostics_channel from "node:diagnostics_channel";
 
 const channels: Record<string | symbol, diagnostics_channel.Channel> = {};
@@ -11,9 +10,11 @@ export class Channel<StoreType, ContextType>
   readonly __unenv__ = true;
 
   name: diagnostics_channel.Channel["name"];
+
   get hasSubscribers() {
     return this._subscribers.length > 0;
   }
+
   _subscribers: diagnostics_channel.ChannelListener[];
 
   constructor(name: diagnostics_channel.Channel["name"]) {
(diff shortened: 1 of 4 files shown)


src/runtime/node/diagnostics_channel/index.ts
```diff
@@ -0,0 +1,182 @@
+import noop from "../../mock/noop";
+import diagnostics_channel from "node:diagnostics_channel";
```
IgorMinar, line 2: hmm.. I'm not sure how you intended this to work but I *think* this import will point back at this file resulting in an infinite loop.

IgorMinar (commented):
I might be missing something here, but it seems that this PR is unnecessary for Cloudflare. This is because we already implement diagnostics_channel fully in workerd and we marked this module as such in Cloudflare preset: https://github.com/unjs/unenv/blob/dc99bf8bc86bce3bea9dff2f21bea7c99ca22306/src/presets/cloudflare.ts#L16

I guess for completeness of unenv, a polyfill like this would be useful for other environments, but the implementation of such polyfill can't rely on imports from `node:diagnostics_channel` as it's done in this PR:

https://github.com/unjs/unenv/pull/200/files#diff-0f9c42639076225271d82fff8ce82ff7db60dc4c71532558680675baac3be2f3R2

@jculvey can you please help clarify the intention here?

jculvey:
> I might be missing something here, but it seems that this PR is unnecessary for Cloudflare.

aww shucks 🤦 .

I overlooked the fact that this was included in `workerd`. My thought process here was that I started working on `dgram`, and realized it had a dependency on `diagnostics_channel`, so I decided to pick that up first thinking it needed to be done.

🤷

src/runtime/node/diagnostics_channel/index.ts
jculvey: Yup, this was supposted to be `import type diagnostics_channel from "node:diagnostics_channel";`, which I've updated it to.

IgorMinar:
hahaha.. I see. oops. 🤣 

well, let's have @pi0 take a look at this one and see if he likes it. I think it's a good addition to nodeless preset, but as we know it won't be used with workerd.

src/runtime/node/diagnostics_channel/index.ts
```diff
@@ -0,0 +1,182 @@
+import noop from "../../mock/noop";
+import type diagnostics_channel from "node:diagnostics_channel";
+
+const channels: Record<string | symbol, diagnostics_channel.Channel> = {};
+
+export class Channel<StoreType, ContextType>
```
pi0, line 6: It would be nicer if we create `channel.ts`

src/runtime/node/diagnostics_channel/index.ts
```diff
@@ -0,0 +1,182 @@
...
+};
+
+export const unsubscribe: typeof diagnostics_channel.unsubscribe = function (
+  name,
+  onMessage,
+) {
+  return (channel(name) as Channel<unknown, object>).unsubscribe(onMessage);
+};
+
+class TracingChannel<StoreType = unknown, ContextType extends object = object>
```
pi0, line 74: this could also be in separate file

IgorMinar (approved):
LGTM

src/runtime/node/diagnostics_channel/channel.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  publish(message: unknown): void {
+    for (const subscriber of this._subscribers) {
+      subscriber(message, this.name);
+    }
+  }
+
+  bindStore() {}
+  unbindStore() {}
+  runStores() {}
```
pi0, line 46: Maybe throwing explicit error in these might make more sense? (otherwise polyfill would make silent issues)

pi0 (approved):

PR #201, opened 2024-05-24
feat: polyfill `node:constants` module

jculvey:
Replaces the current auto-mocking of 'constants' to support destructured ESM imports and allow for functional polyfill coverage in the future.

### ❓ Type of change

### 📚 Description

Polyfills the `node:constants` module. Also relocates the `os` constants to this new module and re-exports them from `os`.

### 📝 Checklist


src/runtime/node/constants/index.ts
```diff
@@ -0,0 +1,16 @@
+// Mostly taken from: https://github.com/nodejs/node/blob/main/typings/internalBinding/constants.d.ts
+
+import constants from "node:constants";
```
IgorMinar, line 3: did you mean `type`?

```suggestion
import type constants from "node:constants";
```

IgorMinar (approved):
This looks reasonable to me assuming these values match the real implementation. Could you verify the match against the api scan in the matrix app with values serialized as string/number literals?

src/runtime/node/constants/crypto.ts
```diff
@@ -0,0 +1,58 @@
+export const OPENSSL_VERSION_NUMBER = 269_488_319;
```
pi0, line 1: Honestly, i'm not sure about these hardcoded versions. 

If any library depends on them for a minimal version for specific behavior, we are kinda giving false information to it that _might_ end-up into a chain of issues or possible vulnerabilities. 

I would set SSL versions to `0` to reduce such risks

IgorMinar: Good catch @pi0. `0` sounds reasonable to me.

IgorMinar (approved):

pi0 (approved):
Thanks!

PR #204, opened 2024-05-24
feat: polyfill `node:domain` module

jculvey:
Replaces the current auto-mocking of 'domain' to support destructured ESM imports and allow for functional polyfill coverage in the future.

### ❓ Type of change

### 📚 Description

Polyfills the `node:domain` module.

### 📝 Checklist

### DIFF
--- a/src/runtime/node/domain/_domain.ts
+++ b/src/runtime/node/domain/_domain.ts
@@ -1,0 +1,18 @@
+import { notImplemented } from "../../_internal/utils";
+import noop from "../../mock/noop";
+import { EventEmitter } from "../events";
+import type domain from "node:domain";
+
+// eslint-disable-next-line unicorn/prefer-event-target
+export class Domain extends EventEmitter implements domain.Domain {
+  readonly __unenv__ = true;
+
+  members = [];
+  add = noop;
+  enter = noop;
+  exit = noop;
+  remove = noop;
+  bind = notImplemented("Domain.bind");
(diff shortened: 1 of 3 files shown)


src/runtime/node/domain/_domain.ts
```diff
@@ -0,0 +1,18 @@
...
+  readonly __unenv__ = true;
+
+  members = [];
+  add = noop;
+  enter = noop;
+  exit = noop;
+  remove = noop;
+  bind = notImplemented("Domain.bind");
+  intercept = notImplemented("Domain.bind");
+  run = notImplemented("Domain.bind");
```
pi0, line 17: This pattern assigns properties in the constructor and this way if some lib extends `Domain` and tries to override this:

```js
class Foo {
  // can be overriden
  // bar () { throw new Error('not implemented' )}

  // cannot be overriden
  bar = () => { throw new Error('not implemented' )}
}

class Foo2 extends Foo {
  bar() {
    console.log('booo')
  }
}

new Foo2().bar() // Throws error still
```

IgorMinar: I didn't realize that properties and methods in JS classes lived in two separate namespaces. That's a bit strange, but likely intentional.

So a better pattern could be:

```js
class Foo { 
  bar () { notImplemented("bar") }
}
```

IgorMinar (approved):
LGTM with tweaks. Thanks James!

pi0 (approved):
Thanks!

PR #209, opened 2024-05-24
feat: polyfill `node:perf_hooks` module and performance polyfill for nodeless

jculvey:
### ❓ Type of change

### 📚 Description

Adds a polyfill for the `node:perf_hooks` module.

Replaces the current auto-mocking of 'perf_hooks' to support destructured ESM imports and allow for functional polyfill coverage in the future.

### 📝 Checklist


src/runtime/node/perf_hooks/index.ts
```diff
@@ -0,0 +1,67 @@
...
+export const monitorEventLoopDelay: typeof perf_hooks.monitorEventLoopDelay =
+  () => mockIntervalHistogram();
+
+export const performance: typeof perf_hooks.performance = mock.__createMock__(
+  "perf_hooks.performance",
+);
+
+// PerformanceNodeTiming is included in the types but doesn't exist in the runtime
+export default <Omit<typeof perf_hooks, "PerformanceNodeTiming">>{
+  Performance,
```
jculvey, line 56: @pi0 I'm noticing that some of these exports are also defined on `global`. Is it ok to add expose  [here](https://github.com/unjs/unenv/blob/main/src/runtime/polyfill/global-this.ts) or is there a better way of doing this?

IgorMinar: I think we should follow the Buffer and process precedence for the globals and expose the side-effect-y files as individual polyfill files, that way environments have fine-grained control over how and when to introduce the globals.

IgorMinar (approved):
Thanks for putting this together James!

If I'm reading this correctly, the default implementation will essentially be noop - that's good for many of the performance APIs but not good enough for `performance.now` which should really return a real value by using Date.now() similar to how we did it here: https://github.com/unjs/unenv/commit/44d29a609facd9350ee0b453746ba3111d7bfc43

I don't know how many of the other performance APIs are critical for good coverage. I suspect that with the exception of `performance.now()` the rest could be mocked out and filled in when/if the need arises.

IgorMinar:
Hmm... digging into this a bit more. I just remembered that we implemented bits and pieces of this in workerd in the past:
- https://github.com/cloudflare/workerd/issues/390
- https://github.com/cloudflare/workerd/pull/603/files

So it would be good if the polyfill could try to preserve the native `globalThis.performance` if it already exists in the runtime and only monkey patch the native object with methods and properties if the object doesn't have these APIs already defined. In other words, could we preserve workerd's performance.now method, and patch onto this object other methods like `mark` if they are not already defined?

@jsnell - this is an interesting one, and very similar to `globalThis.process` where we would like to keep the original native implementation if present, and fill in any missing APIs with mocks and thus creating a hybrid polyfill at an object instance level.

Ideally we build these unenv polyfills in a way where if/when we add more API surface to workerd's `process` or `performance` APIs, the polyfill will simply fill in less and less of the surface. All of this should be dynamic and guided via feature detection in unenv polyfills.

src/runtime/polyfill/performance.ts
```diff
@@ -0,0 +1,25 @@
+import {
+  performance,
+  Performance,
+  PerformanceEntry,
+  PerformanceMark,
+  PerformanceMeasure,
+  PerformanceObserver,
+  PerformanceObserverEntryList,
+  PerformanceResourceTiming,
+} from "../node/perf_hooks/index";
```
pi0, line 10: Most of these APIs are Web APIs supported by Node.js too. For a generic polyfill/shim, I would suggest that we start implementing them in `src/web/*` ([Performance](https://developer.mozilla.org/en-US/docs/Web/API/Performance)). 

If there are slight differences or more Node.js-specific exports we can override them via `node/perf_hooks` but even if identical, a clear export would keep it more future proof because other than Node.js, there would be JS runtimes that need Web standard impl of performances API.

jculvey: ok I took a swing at restructuring, lmk if this is what you had in mind.

I think something is a little off though now with the `web` entrypoint, because although this seems to build fine it's failing in my entrypoint with `[ERROR] Could not resolve "../../web/performance"`. It could be an issue with the way that our test harness is using esbuild.

jculvey: Ah ok perfect! 

Yeah I think this will probably solve the issues 👍

package.json
```diff
@@ -56,4 +56,4 @@
     "wrangler": "^3.57.1"
   },
   "packageManager": "pnpm@9.1.2"
-}
\ No newline at end of file
+}
```
pi0, line 59:
```suggestion
}

```

src/runtime/node/perf_hooks/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+}
+
+export const createHistogram: typeof perf_hooks.createHistogram = () =>
+  mockRecordableHistogram();
+export const monitorEventLoopDelay: typeof perf_hooks.monitorEventLoopDelay =
+  () => mockIntervalHistogram();
+
+// PerformanceNodeTiming is included in the types but doesn't exist in the runtime
+export default <Omit<typeof perf_hooks, "PerformanceNodeTiming">>{
```
pi0, line 36: Checking with Node 20.12.2 REPL, i get a named export `performance` with `nodeTiming` instance:

image

pi0: well that should be fine considering `performance` is a deep mock now!

pi0 (approved):

src/runtime/node/perf_hooks/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const mockRecordableHistogram = () =>
+  Object.create({
+    min: 9_223_372_036_854_776_000,
+    max: 0,
+    mean: null,
+    exceeds: 0,
+    stddev: null,
+    count: 0,
+    percentile: {
```
pi0: just checked locally not sure why types pass but checking in runtime, `percentile` is a function (PS: there are more)

pushed impl in 3ce186d3ee2b5126f86a9768dc3d87932a370988 in the same PR. feel free to amend or update it.

jculvey: Awesome, thanks! I probably just made a typo and confused `percentile` with `percentiles`. I'm also a little surprised typecheck didn't catch this. 

I pushed a commit that makes a base `Histogram` class that both `IntervalHistogram` and `RecordableHistogram` derive from. Let me know if that works or if there's some consideration for not using a base class.

Also just curious, where does `buckets` come from? I don't see that on either type of histogram in node repl.

pi0: Thanks & nice find (buckets: 3AM coding and trusting Copilot auto-completion 😆)

pi0 (approved):

pi0:
Thanks @jculvey and @IgorMinar again.

I've rebased this PR with partial web implementation of performances API, and extending the Node.js additions (some are mainly for type mismatches..). See #213 and #214.

PR #210, opened 2024-05-24
feat: polyfill `node:punycode` module

jculvey:
Replaces the current auto-mocking of 'punycode' to support destructured ESM imports and allow for functional polyfill coverage in the future.

### ❓ Type of change

### 📚 Description

Note that the node `punycode` module is [deprecated](https://nodejs.org/api/punycode.html). This change introduces the suggested `punycode.js` [library](https://github.com/mathiasbynens/punycode.js) as a drop-in replacement similar to how we use `pathe`.

### 📝 Checklist


package.json
```diff
@@ -42,10 +42,12 @@
     "defu": "^6.1.4",
     "mime": "^3.0.0",
     "node-fetch-native": "^1.6.4",
-    "pathe": "^1.1.2"
+    "pathe": "^1.1.2",
+    "punycode": "^2.3.1"
```
pi0, line 46: Here is a well-tested and optimized version of punycode that we can inline in ununv: 

https://github.com/unjs/ufo/blob/main/src/punycode.ts

since ufo is part of unjs ecosystem, we could also export utils from ufo directly because this way most of unjs powered builds (including nitro and nuxt) would reuse same bits and save bundle. PR welcome for this but in the meantime, inlining would be better than introducing this dep.

src/runtime/node/punycode/index.ts
```diff
@@ -0,0 +1,8 @@
+import type punycode from "node:punycode";
+
+import * as _punycode from "punycode/";
```
IgorMinar, line 3: is the `/` suffix required here?

jculvey: Unfortunately, yes, since `punycode` is a node built-in module :(

package.json
jculvey: Sounds good. Just want to clarify a few things:

- When you say inline -- do you mean actually inline by copy pasting the source of that file into unenv, or just subbing `ufo` for the `punycode.js` dependency here and re-exporting the relevant functions
- Does `ufo` provide sufficient coverage? I see `encodeHost` uses the `toASCII` function from the punycode file, but I don't see anything for decoding, and none of the `ufo` decode functions seem to be able to decode punycode urls (just testing with the ones from the node docs, ex `punycode.decode('--dqo34k'); // '☃-⌘' `)

jculvey: @pi0 lmk what you'd like to see done here. I'm still not sure what to do about `decode` and `ucs2` if we switch to `ufo` since it looks like there isn't sufficient coverage.

pi0: Added inlined version of punycode with minimal changes and esm compact fixes let's see later how we can minimize it (with proper testing)

pi0 (approved):

PR #211, opened 2024-05-28
feat: polyfill `node:querystring` module

jculvey:
Replaces the current auto-mocking of 'querystring' to support destructured ESM imports and allow for functional polyfill coverage in the future.

### ❓ Type of change

### 📚 Description

Uses `ufo` to add a functional polyfill for the `querystring` module.

### 📝 Checklist

### DIFF
--- a/src/runtime/node/querystring/index.ts
+++ b/src/runtime/node/querystring/index.ts
@@ -4,24 +4,33 @@ import { parseQuery, stringifyQuery } from "ufo";
 import type { QueryObject } from "ufo";
 import type querystring from "node:querystring";
 
-const _stringify: typeof querystring.stringify = function (obj) {
-  return stringifyQuery(obj as QueryObject);
-};
-const _parse: typeof querystring.parse = function (obj) {
+export const parse: typeof querystring.parse = function (
+  obj,
+  _sep,
+  _eq,
+  _options,
+) {
   return parseQuery(obj);
 };
+export const decode: typeof querystring.decode = parse;
 
-export const decode: typeof querystring.decode = _parse;
-export const parse: typeof querystring.parse = _parse;
-export const encode: typeof querystring.encode = _stringify;
-export const stringify: typeof querystring.stringify = _stringify;
+export const stringify: typeof querystring.stringify = function (
+  obj,
+  _sep,
+  _eq,
+  _options,
+) {
+  return stringifyQuery(obj as QueryObject);
+};
+export const encode: typeof querystring.encode = stringify;
 
 // eslint-disable-next-line prefer-const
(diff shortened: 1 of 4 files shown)


src/runtime/node/querystring/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  decode as _decode,
+} from "ufo";
+import type { QueryObject } from "ufo";
+import type querystring from "node:querystring";
+import { notImplemented } from "src/runtime/_internal/utils";
+
+const _stringify: typeof querystring.stringify = function (obj) {
+  return stringifyQuery(obj as QueryObject);
+};
+export const decode: typeof querystring.decode = parseQuery;
```
pi0, line 14: While we don't respect/support all params (sep and eq), it might be more clear that we create a wrapper function (this way is also more runtime safe, the second arg `sep` won't be accidentally passed as second arg of `parseQuery`)

src/runtime/node/querystring/index.ts
```diff
@@ -0,0 +1,32 @@
...
+export const encode: typeof querystring.encode = _stringify;
+export const escape: typeof querystring.escape = _encode;
+export const parse: typeof querystring.parse = parseQuery;
+export const stringify: typeof querystring.stringify = _stringify;
+export const unescape: typeof querystring.unescape = _decode;
+export const unescapeBuffer: typeof querystring.unescape = notImplemented(
```
pi0, line 20: > It is exported primarily to allow application code to provide a replacement decoding implementation if necessary by assigning querystring.unescape to an alternative function.

+ 

> By default, the querystring.unescape() method will attempt to use the JavaScript built-in decodeURIComponent() method to decode. If that fails, a safer equivalent that does not throw on malformed URLs will be used.

we might use something like `let unescapeBuffer = decodeURIComponent` perhaps to be more close to impl?

jculvey: Just to clarify, do you think I should use `encodeURIComponent` and `decodeURIComponent` for `escape` and `unescape`, respectively? 

For `unescapeBuffer` should I just `decoreURIComponent` but then return it as a buffer?

pi0: I need to study more but look at this interesting code ([https://github.com/nodejs/node/blob/1de215e285bf8afebe0b9b1908009399e3b1f9e2/lib/querystring.js#L84]) not sure how it can be faster than built-in `encodeURIComponent` but that's what node more precisely do 😆  

(but in general yeah i was mainly thinking to use `let` to be mutable and a close default value not even probably bother to use it in other of our utils since might need ufo changes)

pi0 (approved):

pi0 (approved):

PR #216, opened 2024-05-29
feat(node:readline): polyfill `node:readline` module

jculvey:
Replaces the current auto-mocking of 'readline' to support destructured ESM imports and allow for functional polyfill coverage in the future.

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/runtime/node/readline/index.ts
+++ b/src/runtime/node/readline/index.ts
@@ -1,6 +1,6 @@
 // https://nodejs.org/api/readline.html#readlineclearlinestream-dir-callback
 import type readline from "node:readline";
-import noop from "src/runtime/mock/noop";
+import noop from "../../mock/noop";
 import promises from "./promises";
 import { Interface } from "./interface";
 
--- a/src/runtime/node/readline/index.ts
+++ b/src/runtime/node/readline/index.ts
@@ -1,0 +1,28 @@
+// https://nodejs.org/api/readline.html#readlineclearlinestream-dir-callback
+import type readline from "node:readline";
+import noop from "src/runtime/mock/noop";
+import promises from "./promises";
+import { Interface } from "./interface";
+
+export * as promises from "./promises";
+export { Interface } from "./interface";
+
+export const clearLine: typeof readline.clearLine = () => false;
+export const clearScreenDown: typeof readline.clearScreenDown = () => false;
+export const createInterface: typeof readline.createInterface = () =>
+  new Interface();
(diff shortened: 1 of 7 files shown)


src/runtime/node/readline/interface.ts
```diff
@@ -0,0 +1,44 @@
+import type readline from "node:readline";
+import type { Abortable } from "node:events";
+import { EventEmitter } from "../events";
```
IgorMinar, line 3:
```suggestion
import { EventEmitter } from "node:events";
```

src/runtime/node/readline/interface.ts
```diff
@@ -0,0 +1,44 @@
...
+import { createNotImplementedError } from "src/runtime/_internal/utils";
```
IgorMinar, line 4: shouldn't this be a relative path?

src/runtime/node/readline/index.ts
```diff
@@ -0,0 +1,28 @@
+// https://nodejs.org/api/readline.html#readlineclearlinestream-dir-callback
+import type readline from "node:readline";
+import noop from "src/runtime/mock/noop";
```
IgorMinar, line 3: shouldn't this be a relative path? bare/absolute import like this would require build magic, which doesn't seem worth it.

IgorMinar (approved):
a few tweaks needed, but the rest looks great. Thanks James!

src/runtime/node/readline/interface.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+  setPrompt(prompt: string): void {}
+  prompt(preserveCursor?: boolean | undefined): void {}
+  question(query: string, callback: (answer: string) => void): void;
+  question(
+    query: string,
+    options: Abortable,
+    callback: (answer: string) => void,
+  ): void;
+  question(query: unknown, options: unknown, callback?: unknown): void {}
```
pi0, line 23: The callback will remain pending if we don't fail it

src/runtime/node/readline/interface.ts
```diff
@@ -0,0 +1,44 @@
...
+    return {
+      rows: 0,
+      cols: 0,
+    };
+  }
+  pause() {
+    return this;
+  }
+  [Symbol.asyncIterator](): AsyncIterableIterator<string> {
+    throw createNotImplementedError("Interface.asyncIterator");
```
pi0, line 42: or perhaps simple empty array/iterator?

pi0 (commented):
Thanks!

PR #217, opened 2024-05-29
feat(node:timers): polyfill `node:timers` module

jculvey:
Replaces the current auto-mocking of 'timers' to support destructured ESM imports and allow for functional polyfill coverage in the future.

### ❓ Type of change

### 📝 Checklist


src/runtime/node/timers/promises/index.ts
```diff
@@ -0,0 +1,20 @@
+import { notImplementedAsync } from "../../../_internal/utils";
+import type timers from "node:timers/promises";
+import { Scheduler } from "./scheduler";
+
+export const setImmediate: typeof timers.setImmediate = notImplementedAsync(
```
IgorMinar, line 5: I think this makes sense since setImmediate is quite problematic for runtimes to implement so it's native availability is spotty.

We could try to shim it, but it's challenging because setImmediate is something in between queueMicrotask and setTimeout: https://nodejs.org/en/learn/asynchronous-work/event-loop-timers-and-nexttick#setimmediate-vs-settimeout

Some have used `postMessage` to get closer to the real behavior (e.g. [vscode here](https://github.com/microsoft/vscode/blob/e045e3b1b25de81232b7db32747f31bf90d02d13/src/vs/base/common/platform.ts#L193-L193)) but even then there are issues, proving this to be a problematic api to shim: https://github.com/microsoft/vscode/issues/133173

I'm ok throwing for now and leaving a comment in the code about these considerations and come back to it later once the need arises.

IgorMinar (approved):
a few tweaks needed but over LGTM! thank you

src/runtime/node/timers/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  globalThis.clearInterval || noop;
+export const clearTimeout: typeof timers.clearTimeout =
+  globalThis.clearTimeout || noop;
+
+export const setImmediate: typeof timers.setImmediate =
+  globalThis.setImmediate || notImplementedAsync("timers.setImmediate");
+export const setInterval: typeof timers.setInterval =
+  globalThis.setInterval || notImplementedAsync("timers.setInterval");
+export const setTimeout: typeof timers.setTimeout =
+  globalThis.setTimeout || notImplementedAsync("timers.setTimeout");
```
pi0, line 20: I was thinking in js runtimes that timeouts don't make sense, perhaps we could immediately call the function as fallback? (if goal is compatibility, calling is better than throwing IMO)

src/runtime/node/timers/promises/index.ts
pi0: Maybe we could try a proimisified version of default `setImmediate` and fallback to queueMicrotask? If the end-goal is maximum compatibility, fallback behaviors (even if not precise) IMO would be better than throwing errors

IgorMinar: queueMicrotask will result in ordering being different in the polyfilled version. a postMessage implementation is likely the closes we can get to the real thing, unless we actually implement this in `workerd` natively.

src/runtime/node/timers/index.ts
IgorMinar: I understand the desire for getting the code working, but setTimeout is such a core api that turning it from an async to async call will likely cause more harm than good. I think it's highly unlikely that we'll ever see a runtime without `setInterval` and `setTimeout` in practice. So for these particular APIs we might as well drop the notImplementedAsync fallback to save a few bytes.

pi0: Quick search: QuickJS and MuJS don't implement timers by default and such limited situations where runtimes don't have timers honestly make sense to me similar to work that sets `Date.now()` to `0` in ambient contexts.

PS: Let's consider when Node.js targeted libraries are used in build, they often pull in lots of more polyfills neither are bundle optimized, that few bytes is precision errors.

jculvey: What's the verdict here? `setTimeout || throw` or `setTimeout || call function as fallback`?

src/runtime/node/timers/promises/index.ts
pi0: promisified version of setImmediate could be likely used in in libraries to make a delay. Exposing an implementation from native scheduler of workerd runtime would be amazing. In the meantime i suggest to simply promisify the global setImmediate to give basic functionality rather than breaking.

src/runtime/node/timers/index.ts
pi0: Yep, and I think using second is totally safe for workerd at least..

src/runtime/node/timers/promises/index.ts
IgorMinar: we could do that if setImmediate is present, but keep in mind that many runtimes including deno and workerd don't support it.

src/runtime/node/timers/index.ts
IgorMinar: I'm not convinced about calling the fn as a fallback directly, but since won't have any impact on workerd, I'll defer to @pi0's decision here.

src/runtime/node/timers/promises/index.ts
pi0: Yeah it would be anyway not consistent with Node.js core and i belive runtime-specific Node polyfills would be different...

I think unenv could simply try it best to make things working on best-effort basis at best.

Checking Deno 1.44.0 it seems to work:

image

And bun returning promise with stagenly not calling cb or somehow a bug:

image

pi0: Let's defer this discusison

src/runtime/node/timers/index.ts
pi0: Let's defer this until we need

pi0 (approved):
Thanks! (defered promises and timerless runtimes discussions until we hit the cases we need)

PR #218, opened 2024-05-29
feat(node:tls): polyfill `node:tls` module

jculvey:
Replaces the current auto-mocking of `node:tls` to support destructured ESM imports and allow for functional polyfill coverage in the future.

### ❓ Type of change

### 📝 Checklist


src/runtime/node/tls/index.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Server } from "./server";
+import { SecureContext } from "./secure-context";
+import * as constants from "./constants";
+
+export * from "./constants";
+export { TLSSocket } from "./tls-socket";
+export { Server } from "./server";
+export { SecureContext } from "./secure-context";
+
+export const connect: typeof tls.connect = function () {
```
IgorMinar, line 13: these anonymous functions might be a pain in the butt to deal with when debugging. Could you please name them (`function connect ...`) or export them as functions rather than `const`s? Thanks

IgorMinar (approved):
LGTM after tweaks. thank you!

src/runtime/node/tls/index.ts
pi0: Although v8 uses some dark magic to infer function name from variables, i fully agree explicitly naming functions can help 

---

```js
> const x = function() { throw new Error('a') }
undefined
> x()
Uncaught Error: a
    at x (REPL1:1:30)
> 
```

pi0: ~> https://github.com/unjs/unenv/issues/222

src/runtime/node/tls/secure-context.ts
```diff
@@ -0,0 +1,5 @@
+import type tls from "node:tls";
+
+export class SecureContext implements tls.SecureContext {
+  context = undefined;
```
pi0, line 4: Just double checking (since typescript sometimes misses) isn't expected default an empty object?

src/runtime/node/tls/server.ts
```diff
@@ -0,0 +1,10 @@
+import type tls from "node:tls";
+import { notImplemented } from "../../_internal/utils";
+import { Server as _Server } from "node:net";
+
+export class Server extends _Server implements tls.Server {
+  addContext(hostname: string, context: tls.SecureContextOptions) {}
+  setSecureContext(options: tls.SecureContextOptions) {}
+  setTicketKeys(keys: Buffer) {}
+  getTicketKeys = notImplemented("Server.getTicketKeys");
```
pi0, line 9: we should avoid class prop assignment syntax. also, I guess we could simply couple `set/get` using an internal array?

src/runtime/node/tls/tls-socket.ts
```diff
@@ -0,0 +1,64 @@
+import { notImplemented } from "../../_internal/utils";
+import type tls from "node:tls";
+import { Socket } from "node:net";
+
+export class TLSSocket extends Socket implements tls.TLSSocket {
+  exportKeyingMaterial = notImplemented("TLSSocket.exportKeyingMaterial ");
+  getCipher = notImplemented("TLSSocket.getCipher");
+  getPeerCertificate = notImplemented("TLSSocket.getPeerCertificate");
```
pi0, line 8: we should avoid class property assignment

src/runtime/node/tls/server.ts
jculvey: Unfortunately it seems a little more complicated than that. `ticketKeys` is actually a Buffer, and I'm not super sure of the semantics here. FWIW even [bun's implementation](https://github.com/oven-sh/bun/blob/4b8f89cb73ac2b38f1da6a28eb91861eb7222264/src/js/node/tls.ts#L627-L633) throws not implemented.

Fixed the prop assignment syntax though.

src/runtime/node/tls/server.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export class Server extends _Server implements tls.Server {
+  addContext(hostname: string, context: tls.SecureContextOptions) {}
+  setSecureContext(options: tls.SecureContextOptions) {}
+  setTicketKeys(keys: Buffer) {
+    throw createNotImplementedError("Server.setTicketKeys");
+  }
+  getTicketKeys() {
+    throw createNotImplementedError("Server.getTicketKeys");
+    return Buffer.from("");
```
pi0, line 13: Since throw happens, i think return would not be reached

```suggestion
```

src/runtime/node/tls/tls-socket.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export class TLSSocket extends Socket implements tls.TLSSocket {
+  // exportKeyingMaterial = notImplemented("TLSSocket.exportKeyingMaterial");
+  exportKeyingMaterial() {
+    throw createNotImplementedError("TLSSocket.exportKeyingMaterial");
+    return Buffer.from("");
+  }
+  getCipher() {
+    throw createNotImplementedError("TLSSocket.getCipher");
+    return {} as any;
```
pi0, line 13:
```suggestion
```

src/runtime/node/tls/tls-socket.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+  getPeerCertificate() {
+    throw createNotImplementedError("TLSSocket.getPeerCertificate");
+    return {} as any;
```
pi0, line 17:
```suggestion
```

src/runtime/node/tls/tls-socket.ts
```diff
@@ -0,0 +1,71 @@
...
+    return false;
+  }
+  renegotiate(
+    options: {
+      rejectUnauthorized?: boolean | undefined;
+      requestCert?: boolean | undefined;
+    },
+    callback: (err: Error | null) => void,
+  ) {
+    return undefined;
```
pi0, line 58: Callback to call

PR #220, opened 2024-05-29
feat(node:tty): polyfill `node:tty` module

jculvey:
Replaces the current auto-mocking of 'tty' to support destructured ESM imports and allow for functional polyfill coverage in the future.

### ❓ Type of change

### 📝 Checklist


src/runtime/node/tty/write-stream.ts
```diff
@@ -0,0 +1,38 @@
...
+    return 1;
+  }
+  hasColors(count?: number | undefined): boolean;
+  hasColors(env?: object | undefined): boolean;
+  hasColors(count: number, env?: object | undefined): boolean;
+  hasColors(count?: unknown, env?: unknown): boolean {
+    return false;
+  }
+  getWindowSize(): [number, number] {
+    return [0, 0];
```
IgorMinar, line 33: I wonder if 0 could be a problematic value. maybe not.

I tried to see what webcontainers do and they seem to error due to what looks like an internal error: https://stackblitz.com/edit/stackblitz-starters-ch491u?description=Starter%20project%20for%20Node.js,%20a%20JavaScript%20runtime%20built%20on%20Chrome%27s%20V8%20JavaScript%20engine&file=index.js,package.json&title=node.new%20Starter

// cc: @d3lm

IgorMinar (approved):
LGTM

src/runtime/node/tty/read-stream.ts
```diff
@@ -0,0 +1,10 @@
+import type tty from "node:tty";
+import { Socket } from "node:net";
+
+export class ReadStream extends Socket implements tty.ReadStream {
+  isRaw = false;
+  setRawMode(mode: boolean) {
```
pi0, line 6: Since we already keep track of `isRaw`, we can simply update it with method

src/runtime/node/tty/write-stream.ts
```diff
@@ -0,0 +1,38 @@
+import type tty from "node:tty";
+import { Socket } from "node:net";
+
+export class WriteStream extends Socket implements tty.WriteStream {
+  clearLine(dir: tty.Direction, callback?: (() => void) | undefined) {
```
pi0, line 5: Callbacks in this class (clearLine, clearScreenDown, cursorTo, moveCursor) should be called

src/runtime/node/tty/write-stream.ts
pi0: A mock value makes sense indeed if it is used for formatting / line wrapping in a lib. I think 80x24 commonly is used as fallback (i use 25 for OCD reasons haha!)

pi0: i think we can use class props btw to generate value?)

pi0 (approved):
Thanks!

PR #226, opened 2024-05-30
feat(node:vm): polyfill `node:vm` module

jculvey:
Replaces the current auto-mocking of `vm` to support destructured ESM imports and allow for functional polyfill coverage in the future.

### ❓ Type of change

### 📝 Checklist


src/runtime/node/vm/script.ts
```diff
@@ -0,0 +1,16 @@
+import type vm from "node:vm";
+
+export class Script implements vm.Script {
+  runInContext(
+    contextifiedObject: vm.Context,
+    options?: vm.RunningScriptOptions | undefined,
+  ) {}
```
IgorMinar, line 7: I think we should throw here. I can't imagine a scenario where it would be acceptable for the implementation to be a noop.

src/runtime/node/vm/script.ts
```diff
@@ -0,0 +1,16 @@
...
+  runInNewContext(
+    contextObject?: vm.Context | undefined,
+    options?: vm.RunningScriptInNewContextOptions | undefined,
+  ) {}
+  runInThisContext(options?: vm.RunningScriptOptions | undefined) {}
```
IgorMinar, line 12: same as above for all of these...

src/runtime/node/vm/index.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export { Script } from "./script";
+export * as constants from "./constants";
+
+export const compileFunction: typeof vm.compileFunction =
+  function compileFunction() {
+    return noop;
+  };
+export const createContext: typeof vm.createContext = function createContext() {
+  return {};
```
IgorMinar, line 14:
```suggestion
  return Object.create(null);
```

src/runtime/node/vm/index.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+export const isContext: typeof vm.isContext = () => false;
+export const measureMemory: typeof vm.measureMemory = () =>
+  Promise.resolve({
+    total: { jsMemoryEstimate: 0, jsMemoryRange: [1, 2] },
+    WebAssembly: { code: 0, metadata: 0 },
+  });
+export const runInContext: typeof vm.runInContext = noop;
+export const runInNewContext: typeof vm.runInNewContext = noop;
+export const runInThisContext: typeof vm.runInThisContext = noop;
```
IgorMinar, line 27: as mentioned in script.ts - I think these should throw instead.

src/runtime/node/vm/index.ts
```diff
@@ -0,0 +1,42 @@
...
+import type vm from "node:vm";
+import { Script } from "./script";
+import * as constants from "./constants";
+
+export { Script } from "./script";
+export * as constants from "./constants";
+
+export const compileFunction: typeof vm.compileFunction =
+  function compileFunction() {
+    return noop;
```
IgorMinar, line 11: I think that the returned value should be a function that throws when called, or compileFunction itself should throw when called.

src/runtime/node/vm/script.ts
```diff
@@ -0,0 +1,23 @@
+import type vm from "node:vm";
+import { createNotImplementedError } from "src/runtime/_internal/utils";
+
+export class Script implements vm.Script {
```
pi0, line 4: While it wouldn't work in worked, i was thinking maybe we could emulate it partially with `eval`? (or modern base64 with dynamic import also should work https://github.com/unjs/mlly#evalmodule only issue is async)

IgorMinar: Getting the compliance of `eval`'s implementation right will be tricky and in some cases impossible. Is there any prior art we could reuse? 

Maybe we could leave that out of this PR due to the complexity. I suspect that we'll need two implementations anyway - one that is CSP/workerd compliant and doesn't use eval, and one that uses eval.

pi0: Yeah. Let me see how hard it would be totally agree to defer if not easy.

Re CSP compatibility, i think either way we throw an error right? (if unenv completely not implements `eval` shim or the runtime disallows eval)

pi0: ~> https://github.com/unjs/unenv/issues/236

pi0 (approved):
Thanks!

PR #227, opened 2024-05-30
feat(node:wasi): polyfill `node:wasi` module

jculvey:
Replaces the current auto-mocking of `wasi` to support destructured ESM imports and allow for functional polyfill coverage in the future.

### ❓ Type of change

### 📝 Checklist


src/runtime/node/wasi/wasi.ts
```diff
@@ -0,0 +1,12 @@
+import type wasi from "node:wasi";
+
+export class WASI implements wasi.WASI {
```
IgorMinar, line 3: similar to the `vm` module, I think the constructor of WASI should throw as not implemented. I can't think of why noop implementation would help some applications to run - instead we'll see people confused when their apps fail at runtime with very weird downstream errors.

pi0 (approved):

PR #224, opened 2024-05-30
feat(cloudflare): introduce hybrid polyfills for the cloudflare preset

IgorMinar:
This preset provides hybrid polyfills (https://github.com/unjs/unenv/issues/181) but requires special workerd flags.

Due to potential for breaking changes and extra requirements, we created a new preset rather than updating the existing one.

In the next major version of unenv, we should remove the cloudflare preset and rename cloudflareV2 to cloudflare.

### 🔗 Linked issue

https://github.com/unjs/unenv/issues/181

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #230, opened 2024-05-31
feat(node:zlib): polyfill `node:zlib` module

jculvey:
Replaces the current auto-mocking of `zlib` to support destructured ESM imports and allow for functional polyfill coverage in the future.

### ❓ Type of change

### 📝 Checklist


src/runtime/node/zlib/index.ts
```diff
@@ -0,0 +1,109 @@
...
+export const gzip: typeof zlib.gzip = promisifiedBuffer;
+export const gzipSync: typeof zlib.gzipSync = bufferNoop;
+export const inflate: typeof zlib.inflate = promisifiedBuffer;
+export const inflateRaw: typeof zlib.inflateRaw = promisifiedBuffer;
+export const inflateRawSync: typeof zlib.inflateRawSync = bufferNoop;
+export const inflateSync: typeof zlib.inflateSync = bufferNoop;
+export const unzip: typeof zlib.unzip = promisifiedBuffer;
+export const unzipSync: typeof zlib.unzipSync = bufferNoop;
+
+export default <typeof zlib>{
```
IgorMinar, line 69: I started using this new pattern which relies on `satisfies` - unlike a cast like the one here, `satisfies` enforces the shape of the object and complains when you are not satisfying the requirements of the type.

Can you switch to `satisfies` please?

See example: https://github.com/unjs/unenv/pull/224/files#diff-bffdb1a86eff380b9fda777a70b78ddf439f60992878d64c3fc008c8854092bbR40

src/runtime/node/zlib/classes.ts
IgorMinar, line 1: It feels wrong to me to provide this noop/identity implementation as the default. Like in the other recent PR, I can't imagine a program that would work correctly while relying on this polyfill.

It might be better to error in this case and eventually replace this implementation with a spec compliant one.

src/runtime/node/zlib/contants.ts
```diff
@@ -0,0 +1,115 @@
...
+  BROTLI_DECODER_ERROR_DICTIONARY_NOT_SET: -19,
+  BROTLI_DECODER_ERROR_INVALID_ARGUMENTS: -20,
+  BROTLI_DECODER_ERROR_ALLOC_CONTEXT_MODES: -21,
+  BROTLI_DECODER_ERROR_ALLOC_TREE_GROUPS: -22,
+  BROTLI_DECODER_ERROR_ALLOC_CONTEXT_MAP: -25,
+  BROTLI_DECODER_ERROR_ALLOC_RING_BUFFER_1: -26,
+  BROTLI_DECODER_ERROR_ALLOC_RING_BUFFER_2: -27,
+  BROTLI_DECODER_ERROR_ALLOC_BLOCK_TYPE_TREES: -30,
+  BROTLI_DECODER_ERROR_UNREACHABLE: -31,
+};
```
IgorMinar, line 115: could you typecheck this via `satisfies`?

src/runtime/node/zlib/codes.ts
```diff
@@ -0,0 +1,20 @@
...
+  Z_MEM_ERROR: -4,
+  Z_BUF_ERROR: -5,
+  Z_VERSION_ERROR: -6,
+  "-1": "Z_ERRNO",
+  "-2": "Z_STREAM_ERROR",
+  "-3": "Z_DATA_ERROR",
+  "-4": "Z_MEM_ERROR",
+  "-5": "Z_BUF_ERROR",
+  "-6": "Z_VERSION_ERROR",
+};
```
IgorMinar, line 20: could you typecheck this via `satisfies`?

IgorMinar (commented):
I wonder what @pi0 thinks about this implementation. I think I'd rather throw...

src/runtime/node/zlib/index.ts
pi0: i think we can do it as general refactor over all modules consistently. `satisfies` should do better

jculvey: I like the idea of this too. I'm not sure if it's just this particular module, but the `zlib` type has a lot of inconsistencies with the actual runtime implementation.

Let's circle back and do a general refactor to use `satisfies` consistently and file bugs against `@types/node` when needed and patch up the types where not possible.

pi0:
(working on some reworks, methods need to throw error in callback properly to match spec)

pi0:
Updates:
- sync methods use `notImplemented` as before
- async methods use new util that throws error on **callback** and implements promisify
- compression formats moved to separate modules

Followup idea: https://github.com/unjs/unenv/issues/235

PR #247, opened 2024-05-31
feat: support `inject` with `false` values to  override injections

IgorMinar:
Sometimes an environment might want to undo any kind of inject configuration provided by the previous preset, because this environment guaratees that an API is provided natively and doesn't need to be supported via injection.

An example is Cloudflare's workerd, which natively support Blob and global when the nodejs_compat flag is turned on.

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -24,11 +24,14 @@ export function env(...presets: Preset[]): Environment {
     // Inject
     if (preset.inject) {
       for (const global in preset.inject) {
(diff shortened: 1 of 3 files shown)


pi0 (approved):
Nice idea. Only since we are trying to avoid usage of null across codebases, false to disble might be better?

IgorMinar:
I'm fine with `false`. I'll update later today.

jculvey (approved):
lgtm

PR #249, opened 2024-06-05
feat(node, cloudflare): cleanup `globalThis.process` + add hybrid cloudflare impl

IgorMinar:
This PR makes several related changes to how polyfilling of globalThis.process works:

- the `inject` setting in nodeless and cloudflare presets points to side-effect-free implementation
- the `polyfill` setting in nodeless preset is side-effecty and doesn't have any exports
- I refactored the code to use globalThis directly rather than via globaThis.ts as globalThis is now widely supported and doesn't need to be polyfilled
- I introduced src/runtime/node/process.ts following the precedence of src/runtime/node/global.ts
- I improved type-safety of the files I touched

### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/src/runtime/node/process.ts
+++ b/src/runtime/node/process.ts
@@ -1,0 +1,2 @@
+// injectable version of globalThis.process
+export { default } from "./process/index";
--- a/src/runtime/node/process.ts
+++ b/src/runtime/node/process.ts
@@ -1,2 +1,0 @@
-// injectable version of globalThis.process
-export { default } from "./process/index";
--- a/src/runtime/polyfill/process.ts
+++ b/src/runtime/polyfill/process.ts
@@ -1,6 +1,9 @@
-import _process from "../node/process/index";
-import _global from "./global-this";
+import type nodeProcess from "node:process";
+import unenvProcess from "../node/process/index";
 
-_global.process = _global.process || _process;
-
-export default _global.process;
+// TODO: or should we Object.assign into globalThis.process to preserve the identity
+// of the original globalThis.process?
+globalThis.process = {
+  ...unenvProcess,
+  ...globalThis.process,
+} satisfies typeof nodeProcess;
--- a/src/runtime/polyfill/process.ts
+++ b/src/runtime/polyfill/process.ts
@@ -1,9 +1,8 @@
 import type nodeProcess from "node:process";
 import unenvProcess from "../node/process/index";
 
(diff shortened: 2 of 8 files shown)


src/runtime/polyfill/process.ts
```diff
@@ -1,6 +1,9 @@
-import _process from "../node/process/index";
-import _global from "./global-this";
+import type nodeProcess from "node:process";
+import unenvProcess from "../node/process/index";
 
-_global.process = _global.process || _process;
-
-export default _global.process;
+// TODO: or should we Object.assign into globalThis.process to preserve the identity
```
IgorMinar, line 4: @pi0 I'd love your take on this one. I'm torn as both options have pros and cons

src/runtime/node/process.ts
pi0, line 1: I'm starting to feel we might use `node/_globals` namespace for process/global, etc as cleaner namespace wyt?

src/runtime/polyfill/process.ts
pi0: I would go with `Object.assign` too as it preserves constructor 👍🏼

src/runtime/node/process.ts
IgorMinar: _ usually means private, and these files are not private, they are public entry points configured via `inject`.

The convention that I personally settled on was that `node/<dir>` maps to `node:*` modules, and `node/<file>` maps to globals. I don't feel super strongly about keeping this convention, but I do think it's important to avoid `_` to make it clear that these files are public API.

src/runtime/polyfill/process.ts
IgorMinar: ok, sgtm. I'll update, but I think we should file an issue to do a pass through the library to ensure that we follow this consistently everywhere.

src/runtime/node/process.ts
IgorMinar: how about `$` prefix then?

pi0: as in `src/node/$global`? I would suggest to use `node/_global` still since the usage is really for unenv/cloudflare presets to use they have no users so it is really internal to me (+ we use `$` for provider as convention today)

IgorMinar: how about f75fffa662ef471ff79ad39eb9560d3f9a40fddf then?

pi0 (approved):

PR #255, opened 2024-06-12
feat(node, cloudflare): add missing `node:console` exports + add hybrid polyfill for cloudflare

IgorMinar:
**Reference:** #

### DIFF
--- a/src/runtime/node/console/index.ts
+++ b/src/runtime/node/console/index.ts
@@ -1,9 +1,15 @@
 import type console from "node:console";
+import { Writable } from "node:stream";
 import mock from "../../mock/proxy";
 import noop from "../../mock/noop";
 import { notImplemented } from "../../_internal/utils";
 
 const _console = globalThis.console;
+
+// undocumented public APIs
+export const _ignoreErrors: boolean = true;
+export const _stderr: Writable = new Writable();
+export const _stdout: Writable = new Writable();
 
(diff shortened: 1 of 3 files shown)


jculvey (approved):

src/runtime/node/console/index.ts
```diff
@@ -62,4 +68,10 @@ export default <typeof console>{
...
   trace,
   warn,
-};
+} satisfies typeof console;
+
+export { default as _times } from "../../mock/proxy";
+export {
+  default as _stdoutErrorHandler,
+  default as _stderrErrorHandler,
+} from "../../mock/noop";
```
pi0, line 77: Code style: we usually have default export at last

IgorMinar: this is eslint rewriting my code in a way I don't like. should I disable it?

pi0: It shouldn't order let me check locally 👍🏼

pi0 (approved):
❤️

src/runtime/node/console/index.ts
IgorMinar: I think the issue was that I imported `noop` once and then exported it twice - eslint combined the two into a single export, and along the way changed the order. since you kept it as a single export, you were able to move it up.

PR #256, opened 2024-06-12
feat(node:crypto): add missing `constants` export

IgorMinar:
The const were collected from real node versions, and given major churn in the exported values in Node 18, 20, 22 we export a union of these values to increase compatibility.

**Reference:** #


src/runtime/node/crypto/$cloudflare.ts
```diff
@@ -128,14 +128,14 @@ export default {
    */
   Certificate,
   Cipher,
-  // @ts-expect-error undocumented public API
```
IgorMinar, line 131: I don't know why this was needed before but now it's not needed. I suspect that this some kind of typescript bug. 🤷🏻

jculvey (approved):

src/runtime/node/crypto/internal/constants.ts
```diff
@@ -0,0 +1,221 @@
+/* eslint-disable unicorn/no-useless-spread --
+ * disable eslint rules that make this file hard to recreate/update in the future */
+
+const constants = {
+  // collected from Node.js v18.20.3 using:
+  // require('crypto').constants
+  // require('crypto').constants.defaultCipherList
+  ...{
+    OPENSSL_VERSION_NUMBER: 805_306_576,
```
pi0, line 9: (for versions same as other places i think we can explicitly set to 0 to avoid any assumptions)

```suggestion
    OPENSSL_VERSION_NUMBER: 0,
```

src/runtime/node/crypto/internal/constants.ts
```diff
@@ -0,0 +1,221 @@
...
+    SSL_OP_ALLOW_UNSAFE_LEGACY_RENEGOTIATION: 262_144,
+    SSL_OP_CIPHER_SERVER_PREFERENCE: 4_194_304,
+    SSL_OP_CISCO_ANYCONNECT: 32_768,
+    SSL_OP_COOKIE_EXCHANGE: 8192,
+    SSL_OP_CRYPTOPRO_TLSEXT_BUG: 2_147_483_648,
+    SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS: 2048,
+    SSL_OP_LEGACY_SERVER_CONNECT: 4,
+    SSL_OP_NO_COMPRESSION: 131_072,
+    SSL_OP_NO_ENCRYPT_THEN_MAC: 524_288,
+    SSL_OP_NO_QUERY_MTU: 4096,
```
pi0, line 170: Most of these are repeated 4 times i think we san safely merge them to reduce size and still fulfill purpose of polyfill

IgorMinar: there are some differences between node versions, but you are right, payload size matters. I'll dedupe.

src/runtime/node/crypto/internal/constants.ts
IgorMinar: so set all numbers to `0` and strings to `''`?

pi0: I would set them for **versions** 👍🏼 (OP flags, and numbers i guess are safe unless you belive they also might make implicit behavior in polyfills)

IgorMinar:
@pi0 how about d2398d5? PTAL

pi0 (approved):

Issue #92, opened 2023-04-10
Implement `node:https` polyfill


pi0:
It would be similar to [node/http](https://github.com/unjs/unenv/tree/v1.3.0/src/runtime/node/http). We might also reuse things like constants.

Currently `node:https` is mocked and makes issues like #86

PR #257, opened 2024-06-12
feat(node, cloudflare): fill in api surface for `node:perf_hooks`

IgorMinar:
- added missing APIs
- created a hybrid polyfill for cloudflare
- updated injectable config for globalThis.performance

**Reference:** #


src/presets/nodeless.ts
```diff
@@ -89,7 +89,20 @@ const nodeless: Preset & { alias: Map<string, string> } = {
     global: "unenv/runtime/node/_global",
     process: "unenv/runtime/node/process",
     Buffer: ["buffer", "Buffer"],
-    performance: "unenv/runtime/polyfill/performance",
+    performance: ["perf_hooks", "performance"],
+
+    // TODO: for some reason the following Performance* injectables are not honored
+    //       figure out why... is this some kind of ESBuild bug?
```
IgorMinar, line 95: @pi0 this one really puzzles me. I have no idea what's wrong here. In my testing `globalThis.performance` works as expected but `globalThis.Performance` (capital P) is `undefined`.

Even more oddly, if I replace line #92 with `performance: ["perf_hooks", "Performance"]`, my ESBuild setup replaces `globalThis.performance` with `node:perf_hooks#Performance` as instructed.

So it's as if something about the left-hand-side `Performance` caused the injection to not work. Very bizarre.

Any ideas what could be causing this?

IgorMinar: This is the only thing that remains unresolved in this PR as far as I know and we should probably get to the bottom of it before merging, because otherwise we regress on `globalThis.Performance*` APIs compared to `main` branch.

src/runtime/node/perf_hooks/internal/constants.ts
```diff
@@ -1,15 +1,154 @@
...
+
+// captured from Node.js v22.3.0 using
+// Object.getOwnPropertyDescriptors(require('perf_hooks').constants)
+Object.defineProperties(constants, {
+  NODE_PERFORMANCE_GC_MAJOR: {
+    value: 4,
+    writable: false,
+    enumerable: true,
+    configurable: false,
+  },
```
pi0, line 15: Wondering if an `Object.freeze()` would result in almost identical behavior to reduce size

src/runtime/node/perf_hooks/internal/performance.ts
```diff
@@ -44,6 +67,22 @@ export const Performance = class Performance
...
+    global: object,
+    cacheMode: string,
+    bodyInfo: object,
+    responseStatus: number,
+    deliveryType?: string,
+  ): PerformanceResourceTiming {
+    // TODO: create a new PerformanceResourceTiming entry
+    // so that peformance.getEntries, getEntriesByName, and getEntriesByType return it
+    // see: https://nodejs.org/api/perf_hooks.html#performancemarkresourcetimingtiminginfo-requestedurl-initiatortype-global-cachemode-bodyinfo-responsestatus-deliverytype
+    return mock;
```
pi0, line 83: We have `_PerformanceResourceTiming` class perhaps we can return an instance of it?

src/presets/nodeless.ts
pi0: Nitro uses rollup plugin for injection not sure if it might be a direct issue with unenv...

pi0: Also we have https://github.com/IgorMinar/unenv/blob/4104cad8e86a33851ed2f6d18ae5f27dbedb726f/src/presets/nodeless.ts#L111 - other than explicitly disabled state of cloudflare for nitro i would always globally polyfill. Bundlers can always miss this kind of injections (even if today work, i seen it many times)

IgorMinar: We currently don't use any of the `polyfills` in our wrangler integration and rely solely on `injection`. I'd rather risk the bundler missing some call sites than always bringing in these polyfills and bloating the bundle size.

Unlike the usual usage, [our injection implementation in Wrangler](https://github.com/cloudflare/workers-sdk/blob/874da0e79de18d3661963f8d617209a1bd7816bf/packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts#L114-L188) is side-effecty, so it takes only one successful injection for the global to be polyfilled. That way if a bundler misses a call site, we can have users write noop code in their app that is not missed and its side-effect will bring in the polyfill for all call sites. I think this tree-shakable design is a better trade off between accuracy and size.

IgorMinar: In any case, I'll play with this a bit more today to see if I can figure out what's going on.

src/runtime/node/perf_hooks/internal/constants.ts
IgorMinar: many of the values are non-enumerable and that's why I went with `defineProperties`, but you are right, this can be written in a more size-efficient way. I'll redo.

src/presets/nodeless.ts
pi0: (thinking would it make sense unenv have an explicit `wrangler` preset?)

IgorMinar: I'm not sure. I don't think it's strictly necessary. The current cloudflare preset should work with wrangler and outside of wrangler. The only special thing about the preset is that `inject` and `polyfills` fields are redundant and you can chose to use one or the other or both.

If you want the tree-shakable design that wrangler uses, then you'll need to duplicate some of the code from wrangler and adjust it to whatever bundler you are using (if it's not esbuild already). This is however optional and not a requirement for using the cloudflare preset.

src/runtime/node/perf_hooks/internal/performance.ts
IgorMinar: addressed in c5345ad8c914fec4a55ddab39e15d605dbfb7480

src/presets/nodeless.ts
pi0: Was thinking about next steps (preset test matrix), and mainly if it will give a safer escape hatch for unseen differences in the future (that nitro and wrangler need different behavior). But only a suggestion.

src/runtime/node/perf_hooks/internal/constants.ts
IgorMinar: how about this 29720492bf1ee4ddfaee7eb1e1e438297da398cc - I think this is as small as it gets while staying API surface compliant with Node.

src/presets/nodeless.ts
IgorMinar: I got to the bottom of the original issue. There was an issue in the code in wrangler that I fixed in https://github.com/cloudflare/workers-sdk/pull/6025. So I think this PR is good to go now, I'll remove this TODO.

IgorMinar: > Was thinking about next steps (preset test matrix we need to test cloudflare twice now with and without polyfill - inject only), and mainly if it will give a safer escape hatch for unseen differences in the future (that nitro and wrangler need different behavior). But only a suggestion.

I think that rather than creating a separate preset for wrangler, we should more explicitly document how to integrate unenv with various bundlers. It took me quite a while to figure out how to take the unenv preset and use it with ESBuild which wrangler uses, and the integration is not trivial at all. I can imagine a future where the [ESBuild plugin embedded in wrangler](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts) is a standalone npm package that anyone can use to take an unenv preset and use it with esbuild.

IgorMinar:
@pi0 I believe this PR is good to go. PTAL

src/presets/nodeless.ts
pi0: Yeah same for rolluo integration in Nitro which could be externalized.

pi0: Moved to https://github.com/unjs/unenv/issues/262, https://github.com/unjs/unenv/issues/263 for furthor discussions.

src/presets/nodeless.ts
```diff
@@ -89,7 +89,17 @@ const nodeless: Preset & { alias: Map<string, string> } = {
...
+    Performance: ["perf_hooks", "Performance"],
+    PerformanceEntry: ["perf_hooks", "PerformanceEntry"],
+    PerformanceMark: ["perf_hooks", "PerformanceMark"],
+    PerformanceMeasure: ["perf_hooks", "PerformanceMeasure"],
+    PerformanceObserver: ["perf_hooks", "PerformanceObserver"],
+    PerformanceObserverEntryList: [
+      "perf_hooks",
+      "PerformanceObserverEntryList",
+    ],
+    PerformanceResourceTiming: ["perf_hooks", "PerformanceResourceTiming"],
```
pi0, line 102: To be honest, I'm still not sure about this part. Initial injections in nodeless (`global`, `process`, and `Buffer`) are all Node.js specific global which were safe to inject globally.

[Performance API](https://developer.mozilla.org/en-US/docs/Web/API/Performance) is a Web standard on the other hand, is not. By adding this injections we are effectively **overriding any native global implementation** in user-code.

The issue above is not easily solvable because Node.js variation of implementation needs classes with Constructor support.

I think this should be an explicit opt-in. `wrangler` preset (#262) is a good candidate because i suppose you also require explicit (Node-compat) from users to enable this and then things make sense.

pi0: (i would suggest to revert this part and split to another PR to unblock)

IgorMinar: > [Performance API](https://developer.mozilla.org/en-US/docs/Web/API/Performance) is a Web standard on the other hand. By adding this injections we are effectively overriding any native global implementation in user-code to unenv/node.

You make a really good point, Pooya! It's funny how we sometimes get caught up in not seeing the obvious.

I think the right thing here is to move these injects into the cloudflare preset because workerd doesn't offer Performance* APIs.

I'm not yet convinced that we need a dedicated wrangler preset. I'm hoping we really don't have to do that.

Would you be ok with me updating the PR by moving these injects into `preset/cloudflare`?

pi0: I want to use same cloudflare preset in Nitro builds as well and users would face similar implicit opt-in to Node.js variant of `Performance*` variants if we do this for them...

IgorMinar: I see what you mean. The issue you are bringing up is that Node's Performance* API is not 100% compatible with Web's Performance* and if someone relies on globals we'd be making the choice for them, which might not be good. This is quite messy. How big is the difference between the two?

@jasnell as a node core contributor, have you seen this become a problem in the node community? how do people deal with these compat issues?

pi0: - Current web impl: https://github.com/unjs/unenv/tree/main/src/runtime/web/performance
- Current node impl (extending web): https://github.com/unjs/unenv/blob/main/src/runtime/node/perf_hooks/internal/performance.ts

Difference is:
- In node impl, we always implement class and won't fall back to `globalThis.*` natives if available 
- Node.js classes have additional props

The first is solvable but I'm more worried about the design decisions we hare making for unenv. For me unless explicitly we are adding "node.js" compatibility, we should strictly stick with Web standards spec. Only for wrangler usage that uses unenv as Node.js compatibility layer it clearly makes sense but for generic polyfill, it breaks the concept.

I am thinking to expose new API from unenv that we can have flags like `nodeCompat` to make difference behavior.


---

In the meantime, let's move these to cloudflare as you suggested this PR seems unnecessarily being blocked.

jasnell: we do have `performance.now()` and `performance.timeOrigin` in workers. The impl in Node.js definitely has a bunch of Node.js specific stuff. It hasn't been too much of a problem in the ecosystem but it definitely warrants caution.

IgorMinar:
@pi0 I'll get back to this one within a few days. I still care, just busy with other things.

pi0:
No rush, thanks for the update @IgorMinar

pi0:
Dear @IgorMinar i have rebased your work and moved the injections directly to cloudflare preset to move forward. please feel free to amend if any change needed.

PR #259, opened 2024-06-12
fix(node:url): add `Url` and `resolveObject`

pi0:
Add and use missing `Url` (URL + path, querty, ...) props + `resolveObject` which is not typed but exsits in runtime

### DIFF
--- a/src/runtime/node/url/index.ts
+++ b/src/runtime/node/url/index.ts
@@ -1,20 +1,41 @@
 // https://nodejs.org/api/url.html
-import type url from "node:url";
+import type nodeUrl from "node:url";
+import { ParsedUrlQuery } from "node:querystring";
 
 // TODO: https://github.com/unjs/unenv/issues/65
-export const URL = globalThis.URL as typeof url.URL;
+export const URL = globalThis.URL as typeof nodeUrl.URL;
 
-export const URLSearchParams = globalThis.URLSearchParams;
+export const URLSearchParams =
+  globalThis.URLSearchParams as typeof nodeUrl.URLSearchParams;
+
+// Output of `url.parse`
+export class Url extends URL implements nodeUrl.Url {
+  constructor(input: string, base?: string | URL) {
+    super(input, base);
+  }
+  get path() {
+    return this.pathname;
+  }
+  get query() {
+    return this.search;
+  }
+  get slashes() {
+    return this.pathname.startsWith("/");
+  }
+  get auth() {
+    return this.username + (this.password ? ":" + this.password : "");
+  }
+}
 
 // https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost
 // TODO
-export const parse = <typeof url.parse>(
+export const parse = <typeof nodeUrl.parse>(
   function (
     urlString,
     parseQueryString?,
     slashesDenoteHost?,
-  ): URL | url.UrlWithStringQuery {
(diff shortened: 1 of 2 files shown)


src/runtime/node/url/index.ts
```diff
@@ -84,33 +107,46 @@ export const format = <typeof url.format>(
...
+  relative: string,
+) {
+  if (!source) return relative;
+  // Node.js logic: return urlParse(source, false, true).resolveObject(relative);
+  return new URL(relative, source).toString();
+};
+
+export default {
   URL,
+  // @ts-expect-error
```
IgorMinar, line 148: can you add a reason for the exception so that we know when it's safe to remove it? is this a typings issue?

pi0: Yeah and i guess it would be never typed since is deperecated...

src/runtime/node/url/index.ts
```diff
@@ -33,29 +54,31 @@ export const resolve = <typeof url.resolve>function (from, to) {
...
+        : "",
+    };
+  }
+);
 
 // https://nodejs.org/api/url.html#url_url_format_urlobject
 // TODO
-export const format = <typeof url.format>(
-  function (urlInput, options?: url.URLFormatOptions) {
+export const format = <typeof nodeUrl.format>(
```
IgorMinar, line 80: could you write this as, which actually typechecks the right-handside as opposed to the original which just casts the value without typechecking?
```suggestion
export const format: typeof nodeUrl.format = (
```

src/runtime/node/url/index.ts
```diff
@@ -1,20 +1,41 @@
...
 // TODO: https://github.com/unjs/unenv/issues/65
-export const URL = globalThis.URL as typeof url.URL;
+export const URL = globalThis.URL as typeof nodeUrl.URL;
 
-export const URLSearchParams = globalThis.URLSearchParams;
+export const URLSearchParams =
+  globalThis.URLSearchParams as typeof nodeUrl.URLSearchParams;
+
+// Output of `url.parse`
+export class Url extends URL implements nodeUrl.Url {
```
IgorMinar, line 12: oof.. this is brave. my understanding is that there are significant differences between how urls are parsed by `URL` and `Url`. Do you think we can get away with this? At minimum we should document that we know this is not a Node.js compliant implementation of `Url`.

IgorMinar (commented):
nice improvement! I left a few suggestions

src/runtime/node/url/index.ts
pi0: they are same (i did it for consistency with rest of file ... which is not best format!)

src/runtime/node/url/index.ts
pi0: We can copy their impl ([here](https://github.com/nodejs/node/blob/main/lib/url.js)) but it will pull-in huge amount of polyfills.

I would suggest iterate over this and see if anything is reported since current state was worse (without adding Node.js specific props)

IgorMinar: yeah. I saw their code. I also noticed that Deno just took the code as is and embedded it. Let's go with your implementation and if it's causing headaches, we can implement this one natively in workerd for CF and reuse the same code in unenv for non-CF usage.

IgorMinar: Can you please add a comment into the code stating that this is could be a problem and we might need to reimplement it?

PR #264, opened 2024-06-18
feat!: rewrite `node:events` with upstream node impl

pi0:
This PR rewrites `node:events` module using upstream Node.js code with class and added type safety.

Previous implementation was another rewrite of older browserify in unenv but adding new functionalities became more troubling.


pi0:
(not tested enough but merging to to test via nightly + followup on other modules that need it)

PR #265, opened 2024-06-18
refactor(node/cluster): reimplement

pi0:
Reimplement node:cluster to improve runtime polyfill and ESM coverage


pi0:
@culvey merging to iterate quickly. please ping if anything could be improved.

PR #268, opened 2024-06-18
fix: add missing `node:process` named exports

pi0:
Add missing named exports of `node:process`


IgorMinar:
@pi0  I'm not too keen on this change. I was hoping that we could keep polyfills exportless and side-effecty and use proper modules to power `inject` which requires default export.

The only reason why you needed to make this change is because in some installations the nightly version got hoisted incorrectly and overshadowed the 1.x version. This resulted in a version skew between the built-time and run-time parts of unenv (built-time was on 1.x which pointed at run-time paths that got resolved @ nightly version causing the problem).

I'd rather think about how to fix the version skew problem than try to align the runtime@head to be compatible with every past version of unenv released.

PR #270, opened 2024-06-18
refactor(node): import from `node:` for non extend usages

pi0:
I have checked recent changes for usage for node modules within polyfills.

With explicit import of `node:`, we can allow hybrid polyfills.

With explicit relative import, we can allow polyfills extending each other.

This PR unifies to use `node:` with exception of polyfill classes extending each other -- other than EventEmitter : (constructors like `Readable` are imported from `node:*`)

Relative dependencies summary:

```
fetch => node/http (virtual request)
node/http => node/stream (extends polyfills)
node/net => node/stream (extends polyfills)
node/tls => node/net (extends polyfills)
node/tty => node/net (extends polyfills)
```

### DIFF
--- a/src/runtime/node/dgram/internal/socket.ts
+++ b/src/runtime/node/dgram/internal/socket.ts
@@ -1,4 +1,4 @@
-import { EventEmitter } from "../../events";
+import { EventEmitter } from "node:events";
 import type net from "node:net";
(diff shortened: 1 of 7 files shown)


pi0:
Merging to iterate

Issue #290, opened 2024-08-01
url.parse('/404') throws, unlike in Node.js


mhart:
### Environment

Cloudflare Workers: wrangler 3.64.0 w/ `experimental:nodejs_compat_v2`

### Reproduction

`require('url').parse('/404')`

### Describe the bug

You get `TypeError: Failed to construct 'URL': Invalid URL` (or similar) due to unenv's use of `new URL()` to parse.

In Node.js, with `url.parse('/404')` you get:
```
Url {
  protocol: null,
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: null,
  search: null,
  query: null,
  pathname: '/404',
  path: '/404',
  href: '/404'
}
```

### Additional context

This occurs in Next.js as it passes around "urls" like `/404` and `/500` internally a lot.

The [code comments note divergence is potentially an issue](https://github.com/unjs/unenv/blob/1a799443a5d3f0463531f715e22ef0b1e78b43ba/src/runtime/node/url/index.ts#L11-L13) – and indeed this is a case where it is.

The compat matrix over at https://workers-nodejs-compat-matrix.pages.dev/ should probably be changed to be a "Mismatch" instead of a "Supported" for url.parse due to this (and no doubt other strings exhibit divergence too)

PR #292, opened 2024-08-09
Replace polyfillyed url impl with node-url ecosystem module

jasnell:
Pulls in the `url` polyfill from npm to provide an implementation of `url.parse`/`url.format`/`url.resolve` that has feature parity with Node.js rather than building on the WHATWG URL impl

Refs: https://github.com/unjs/unenv/issues/290

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,8259 @@
+{
+  "name": "unenv",
+  "version": "1.9.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "unenv",
+      "version": "1.9.0",
+      "license": "MIT",
+      "dependencies": {
+        "defu": "^6.1.4",
+        "node-url": "npm:url@^0.11.4",
+        "ohash": "^1.1.3",
+        "pathe": "^1.1.2",
+        "ufo": "^1.5.4"
+      },
+      "devDependencies": {
+        "@types/node": "^20.14.12",
+        "automd": "^0.3.8",
+        "changelogen": "^0.5.5",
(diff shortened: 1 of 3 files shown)


anonrig (commented):
Nice!

mhart:
This doesn't cover the full API though – things like `pathToFileURL` – is there not a polyfill that does cover all these things?

jasnell:
> ... is there not a polyfilly that does cover all these things?

Not that I'm immediately aware of.

pi0:
Thanks for the PR dear @jasnell. General approach we took in unenv was to directly port source from Node.js lib (which i prefer to move forward for this) or either port polyfill libs directly into source tree of unnv as ESM+Typesscript module. `url` npm package seems also a CommonJS-only dependency.

I will try to followup on this PR in coming days.

vicb:
@pi0 any update here?

pi0:
@vicb is it blocking you? (if yes, please kindly share reference)

vicb:
It would help with https://github.com/dario-piotrowicz/poc-build-nextjs-app-for-cf-workers/issues/6 - you can see Dario referenced this PR in the issue.

I can make sure it unblocks us before merging this PR.

What is the status of your alternative solution?

Thanks!

pi0:
Thank you for the reference.

IgorMinar:
@pi0 do you have any more concerns with this PR? Could we get this in or could you please elaborate on what makes you hesitant about this change?

pi0:
@IgorMinar (sent DM in discord)

jasnell:
Closing in favor of @pi0's preferred alternative approach.

PR #299, opened 2024-08-29
feat(node): fully implement `node:querystring` and `node:url`

### LINKED ISSUE #290: url.parse('/404') throws, unlike in Node.js

pi0:
fixes #290
alternative of #292

Ported implementation of `node:url` and `node:querystring` from [Node.js v22.7.0](https://github.com/nodejs/node/tree/v22.7.0/lib).

Remarks: 
- Process was semi-automated and I tried to not touch implementation as much as possible
- Type checks were also added and passed, although some type safety (mostly for nullable values) had been bypassed trusting upstream implementation is safe
- Ported `tests/parallel/test-{querystring,url}*` tests which pass as expected
  - [ ] Invalid IDNA seems to be uncaught. possibly related to `punycode` behavior not matching native format binding
  - [ ] Whatwg URL format with unicode+auth options fails


src/runtime/node/querystring/internal/querystring.ts
```diff
@@ -0,0 +1,113 @@
...
+      hexTable[0x80 | (c & 0x3f)];
+  }
+  if (lastPos === 0) return str;
+  if (lastPos < len) return out + String.prototype.slice.call(str, lastPos);
+  return out;
+}
+
+export { encodeStr };
+export { hexTable };
+export { isHexTable };
```
anonrig, line 113:
```suggestion
export { encodeStr, hexTable, isHexTable };
```

anonrig (approved):

src/runtime/node/querystring/internal/querystring.ts
pi0: Thanks. btw this is auto generated code from unjs/nstdlib (private but you should have access)

pi0:
dear @IgorMinar i noticed you added yourself for review. Do you have any specific concerns about this PR?

I think for next step of adopting hybrid support for workerd, we can try with another PR. (and i have two different methods i like to discuss with your team how we can do it).

pi0:
(merging to test against nightly channel and iterate)

PR #301, opened 2024-09-02
feat(cloudflare): implement `node:module` as an hybrid polyfill

vicb:
Related `workerd` PR: https://github.com/cloudflare/workerd/pull/2636

This PR changes `module` to be an hybrid polyfill and use the workerd available method implementations.

Is there anything else to change/update?

Thanks!

/cc @pi0

### DIFF
--- a/src/runtime/node/module/$cloudflare.ts
+++ b/src/runtime/node/module/$cloudflare.ts
@@ -1,0 +1,88 @@
+// https://nodejs.org/api/module.html
+import type nodeModule from "node:module";
+
+export {
+  Module,
+  SourceMap,
+  _cache,
+  _extensions,
+  _debug,
+  _pathCache,
+  _findPath,
+  _initPaths,
+  _load,
+  _nodeModulePaths,
+  _preloadModules,
+  _resolveFilename,
+  _resolveLookupPaths,
+} from "./index";
+
+import {
+  Module,
+  SourceMap,
+  _cache,
+  _extensions,
+  _debug,
+  _pathCache,
+  _findPath,
+  _initPaths,
+  _load,
+  _nodeModulePaths,
+  _preloadModules,
+  _resolveFilename,
+  _resolveLookupPaths,
+  builtinModules as unenvBuiltinModules,
+  createRequire as unenvCreateRequire,
+  findSourceMap as unenvFindSourceMap,
+  globalPaths as unenvGlobalPaths,
+  isBuiltin as unenvIsBuiltin,
+  register as unenvRegister,
+  runMain as unenvRunMain,
+  syncBuiltinESMExports as unenvSyncBuiltinESMExports,
+  wrap as unenvWrap,
+} from "./index";
+
+const workerdModule = process.getBuiltinModule("node:module");
+
+const resolvedExports = {
+  builtinModules: workerdModule?.builtinModules ?? unenvBuiltinModules,
+  createRequire: { ...unenvCreateRequire, ...workerdModule?.createRequire },
+  findSourceMap: unenvFindSourceMap,
+  globalPaths: unenvGlobalPaths,
+  isBuiltin: workerdModule?.isBuiltin ?? unenvIsBuiltin,
+  register: unenvRegister,
+  runMain: unenvRunMain,
+  syncBuiltinESMExports: unenvSyncBuiltinESMExports,
+  wrap: unenvWrap,
+} satisfies nodeModule;
+
+export const {
+  builtinModules,
+  createRequire,
+  findSourceMap,
+  globalPaths,
+  isBuiltin,
+  register,
+  runMain,
+  syncBuiltinESMExports,
+  wrap,
+} = resolvedExports;
+
+// polyfill missing module API in workerd, while preserving its identity
+Object.assign(resolvedExports, {
+  Module,
+  SourceMap,
+  _cache,
+  _extensions,
+  _debug,
+  _pathCache,
+  _findPath,
+  _initPaths,
+  _load,
+  _nodeModulePaths,
+  _preloadModules,
+  _resolveFilename,
+  _resolveLookupPaths,
+});
+
+export default resolvedExports;
--- a/src/runtime/node/module/$cloudflare.ts
+++ b/src/runtime/node/module/$cloudflare.ts
@@ -61,16 +61,25 @@ export function createRequire(
     return requirePolyfill;
   }
 
-  const requireFn = workerdModule.createRequire(file) as any;
+  const requireFn = workerdModule.createRequire(file) as ReturnType<
+    typeof unenvCreateRequire
+  >;
 
   // Patch properties missing from `workerd`.
-  for (const key of ["resolve", "cache", "extensions", "main"] as const) {
-    if (!(key in requireFn)) {
-      requireFn[key] = requirePolyfill[key];
-    }
+  if (!requireFn.resolve) {
+    requireFn.resolve = requirePolyfill.resolve;
+  }
+  if (!requireFn.cache) {
+    requireFn.cache = requirePolyfill.cache;
+  }
+  if (!requireFn.extensions) {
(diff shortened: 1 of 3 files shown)


vicb:
> Nice start dear @vicb.

Thanks Pooya ;)

I have tried to do something.
The idea is to be able to merge the PR even before the workerd implementation is updated.

I probably got a few things wrong, but I'll learn...

Thanks for the review

src/runtime/node/module/$cloudflare.ts
```diff
@@ -0,0 +1,88 @@
...
+} from "./index";
+
+const workerdModule = process.getBuiltinModule("node:module");
+
+const resolvedExports = {
+  builtinModules: workerdModule?.builtinModules ?? unenvBuiltinModules,
+  createRequire: { ...unenvCreateRequire, ...workerdModule?.createRequire },
+  findSourceMap: unenvFindSourceMap,
+  globalPaths: unenvGlobalPaths,
+  isBuiltin: workerdModule?.isBuiltin ?? unenvIsBuiltin,
```
pi0, line 52: https://github.com/cloudflare/workerd/pull/2636#issuecomment-2324628966

This is more accurate in context of what workerd natively supports but probably breaks compatibility. (many libraries depend on it to know if an import path is external or a known `node:` built-in not if natively supported or not)

src/runtime/node/module/$cloudflare.ts
```diff
@@ -0,0 +1,88 @@
...
+  findSourceMap as unenvFindSourceMap,
+  globalPaths as unenvGlobalPaths,
+  isBuiltin as unenvIsBuiltin,
+  register as unenvRegister,
+  runMain as unenvRunMain,
+  syncBuiltinESMExports as unenvSyncBuiltinESMExports,
+  wrap as unenvWrap,
+} from "./index";
+
+const workerdModule = process.getBuiltinModule("node:module");
```
pi0, line 45: @IgorMinar has to confirm but i guess `process` used to be injected with polyfill)

src/runtime/node/module/$cloudflare.ts
```diff
@@ -0,0 +1,88 @@
...
+  builtinModules,
+  createRequire,
+  findSourceMap,
+  globalPaths,
+  isBuiltin,
+  register,
+  runMain,
+  syncBuiltinESMExports,
+  wrap,
+} = resolvedExports;
```
pi0, line 69: We can refactor it with explicit exports (`export const export_name = (cf or unenv)` and then agreegate default export in single object. This way we don't need additional `Object.assign` overhead too.

src/runtime/node/module/$cloudflare.ts
vicb: We use it at https://github.com/unjs/unenv/blob/70db6f157dde54d1b11f4fa77cee630fb255be49/src/runtime/node/buffer/%24cloudflare.ts#L29

vicb:
Is the last version better?

src/runtime/node/module/$cloudflare.ts
```diff
@@ -0,0 +1,88 @@
...
+  syncBuiltinESMExports,
+  wrap,
+} from "./index";
+
+const workerdModule = process.getBuiltinModule("node:module");
+
+const createRequire: any = workerdModule?.createRequire ?? unenvCreateRequire;
+
+if (workerdModule?.createRequire === undefined) {
+  for (const [key, value] of Object.entries(unenvCreateRequire)) {
```
pi0, line 58: createRequire itself is a method. It's return value is a require function with attached additional methods such as `require.resolv`. Also we need to do this unconditional since the workerd result, does not has all attached functions.

What we can do is something like this:

```js
export function createRequire(...) {
  let requirePolyfill = unenvCreateRequire(...)
  if (!workerdModule?.createRequire) {
    return requirePolyfill
  }
  const require = workerdModule.createRequire(...)
  // Attach missing props from polyfill to require instance
  return require
}
```

src/runtime/node/module/$cloudflare.ts
pi0: Indeed which got me suprised because AFAIK it was problematic for workerd in someplace else there are comments about it.

src/runtime/node/module/$cloudflare.ts
vicb: This should be fixed in the last push

src/runtime/node/module/$cloudflare.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  const require = workerdModule.createRequire(file) as any;
+
+  // Patch properties missing from `workerd`.
+  for (const [key, value] of Object.entries(requirePolyfill)) {
+    if (!(key in require)) {
+      require[key] = value;
+    }
+  }
```
pi0, line 71: The keys we need to attach are known (`resolve: () => {}`, `extensions: {}`, `cache: {}` and `main: undefined`). Instead of a for loop, we can simply set them. this also helps better type safety.

src/runtime/node/module/$cloudflare.ts
```diff
@@ -0,0 +1,99 @@
...
+export function createRequire(
+  file: string,
+): ReturnType<typeof unenvCreateRequire> {
+  const requirePolyfill = unenvCreateRequire(file);
+  if (!workerdModule?.createRequire) {
+    // Use the unenv version of `createRequire` when not supported by `workerd`.
+    return requirePolyfill;
+  }
+
+  const require = workerdModule.createRequire(file) as any;
```
pi0, line 64: I think it might be safer if we call it `requireFn` or something different than `required`. I'm afraid some build transform might wrongly transpile this as actual require.

src/runtime/node/module/$cloudflare.ts
vicb: I thought about that but decided against to support potential added props in a future Node version.
If have a list of property, we will need to update both the polyfill and this file which is easy to miss.

vicb: _I can change if you feel strongly about it_

pi0: I don't think anything new will be added to CJS's require function API.

src/runtime/node/module/$cloudflare.ts
```diff
@@ -0,0 +1,99 @@
...
+    // Use the unenv version of `createRequire` when not supported by `workerd`.
+    return requirePolyfill;
+  }
+
+  const requireFn = workerdModule.createRequire(file) as any;
+
+  // Patch properties missing from `workerd`.
+  for (const key of ["resolve", "cache", "extensions", "main"] as const) {
+    if (!(key in requireFn)) {
+      requireFn[key] = requirePolyfill[key];
```
pi0, line 69: Can you please use explicit lines `if (!requireFn.resolve) ...`. Other than more type safety, it has better perf as we avoid iterator and allocation of a string of arrays..

Also instead of `any` for `requireFn` type we can cast it as require so when assigning typecheck happens,

vicb: Updated, thanks for the reviews

pi0 (approved):
Thanks ❤️ 

LGTM only we should check `process.getBuiltinModule` esbuild issue with @IgorMinar

vicb:
> Thanks ❤️
> 
> LGTM only we should check `process.getBuiltinModule` esbuild issue with @IgorMinar

I think using `process.getBuiltinModule` is ok, it is unpatched:

https://github.com/unjs/unenv/blob/70db6f157dde54d1b11f4fa77cee630fb255be49/src/runtime/node/process/%24cloudflare.ts#L210-L215

_Confirmed that it WAI on a simple wrangler pjt_

src/runtime/node/module/$cloudflare.ts
pi0: https://github.com/unjs/unenv/pull/301#issuecomment-2326151315

PR #302, opened 2024-09-06
feat(cloudflare): use unenv implementation of `setImmediate` for `node:timers`

vicb:
We are exploring using unenv for `setImmediate`.
The desired behavior is run the callback in a `setTimeout(0)` instead of right away.

### DIFF
--- a/src/runtime/node/timers/$cloudflare.ts
+++ b/src/runtime/node/timers/$cloudflare.ts
@@ -1,0 +1,52 @@
+import type timers from "node:timers";
+
+export {
+  _unrefActive,
+  active,
+  clearInterval,
+  clearTimeout,
+  enroll,
+  promises,
+  setInterval,
+  setTimeout,
+  unenroll,
+} from "./index";
+
+export {
+  setImmediateFallback as setImmediate,
+  clearImmediateFallback as clearImmediate,
+} from "./internal/set-immediate";
+
+import {
+  _unrefActive,
+  active,
+  clearInterval,
+  clearTimeout,
+  enroll,
+  promises,
+  setInterval,
+  setTimeout,
+  unenroll,
+} from "./index";
+
+import {
+  setImmediateFallback as setImmediate,
+  clearImmediateFallback as clearImmediate,
+} from "./internal/set-immediate";
+
+globalThis.setImmediate = setImmediate;
+globalThis.clearImmediate = clearImmediate;
+
+export default <typeof timers>{
+  _unrefActive,
+  active,
+  clearImmediate,
+  clearInterval,
+  clearTimeout,
+  enroll,
+  promises,
+  setImmediate,
+  setInterval,
+  setTimeout,
+  unenroll,
+};
--- a/src/runtime/node/timers/$cloudflare.ts
+++ b/src/runtime/node/timers/$cloudflare.ts
@@ -10,6 +10,10 @@ import {
 import { setIntervalFallback } from "./internal/set-interval";
 
 // Always use the polyfill rather than the worked implementation.
+//
+// NOTE:
+// `setImmediate` and `clearImmediate` must be function (re)exports.
+// `export const setImmediate = ...` might cause esbuild to generate invalid code.
 export {
   setImmediateFallback as setImmediate,
   clearImmediateFallback as clearImmediate,
--- a/src/runtime/node/timers/$cloudflare.ts
+++ b/src/runtime/node/timers/$cloudflare.ts
@@ -1,41 +1,36 @@
+import { notImplemented } from "../../_internal/utils";
+import noop from "../../mock/noop";
 import type timers from "node:timers";
+import promises from "./promises";
+import { setTimeoutFallback } from "./internal/set-timeout";
+import {
+  setImmediateFallback as setImmediate,
+  clearImmediateFallback as clearImmediate,
+} from "./internal/set-immediate";
+import { setIntervalFallback } from "./internal/set-interval";
 
-export {
-  _unrefActive,
-  active,
-  clearInterval,
-  clearTimeout,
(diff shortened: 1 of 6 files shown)


src/runtime/node/timers/internal/immediate.ts
```diff
@@ -17,3 +17,31 @@ export class Immediate<TArgs extends any[]> implements NodeJS.Immediate {
...
+
+export class ImmediateTimeoutBased<TArgs extends any[]>
+  implements NodeJS.Immediate
+{
+  _onImmediate: (...args: TArgs) => void;
+  private _timeout: NodeJS.Timeout;
+
+  constructor(callback: (...args: TArgs) => void, args: TArgs) {
+    this._onImmediate = callback;
+    this._timeout = setTimeout(callback, 0, ...args);
```
pi0, line 29: I think you can directly update `Immediate` class implementation to make use of timeout (and we can extract this change as a separate PR)

vicb: So the change would not be cf specific?

pi0: Yes! Using setTimeout seems obviously better choice than immediately calling fn

vicb:
@IgorMinar @pi0 helped me to get https://github.com/unjs/unenv/pull/303 in - it changes the `setImmediate` behavior to use a timeout.

The current behavior is to use `globalThis || unenv`.

This PR would make sure we always use `unenv`.

src/runtime/node/timers/$cloudflare.ts
```diff
@@ -0,0 +1,52 @@
...
+  unenroll,
+} from "./index";
+
+import {
+  setImmediateFallback as setImmediate,
+  clearImmediateFallback as clearImmediate,
+} from "./internal/set-immediate";
+
+globalThis.setImmediate = setImmediate;
+globalThis.clearImmediate = clearImmediate;
```
IgorMinar, line 38: the challenge with this implementation is that this module is now side-effecty which is what we've been trying to avoid in all modules in  unenv except for the `inject` config here: https://github.com/unjs/unenv/blob/f1c0bb8f0feeb22537e15b7f443af352dfb94777/src/presets/cloudflare.ts#L62-L69

could you please remove these two lines and add two new entries to `inject` section of the cloudflare.ts preset?

IgorMinar: you can use `globalThis.console` or `process` as implemented today via `inject` as the pattern to follow. thank you

IgorMinar (changes requested):

vicb:
So...

```js
  inject: {
    setImmediate: ["unenv/runtime/node/timers/$cloudflare", "setImmediate"],
    clearImmediate: ["unenv/runtime/node/timers/$cloudflare", "clearImmediate"],
  },
```

was not working with this code before the PR

```js
// set-immediate.ts

export const clearImmediateFallback = function clearImmediate(...){...};
```

Because the code generated by wrangler was

```js
// At this point clearImmediateFallback === undefined
globalThis.clearImmediate = clearImmediateFallback;

// ...

var setImmediateFallback = function (...) {...};
var clearImmediateFallback = function (...) {...};

// setImmediate is ok
globalThis.setImmediate = setImmediatefFallback;
```

By switching exports to functions, it solves the issue, the generated code is now

```js
// At this point clearImmediateFallback === undefined
globalThis.clearImmediate = clearImmediateFallback;

// ...

function setImmediateFallback(...) {...}
function clearImmediateFallback(...) {...}

// setImmediate is ok
globalThis.setImmediate = setImmediatefFallback;

```

Note: wrangler is using an older version of `esbuild` but updating to the current `0.23.1` still exhibits the same behavior.

IgorMinar (approved):
bah! I'm sorry @vicb, this must have been frustrating to debug. I'm glad that you got to the bottom of it and resolved it. Thank you!

The rest looks great 🧡

src/runtime/node/timers/$cloudflare.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const setTimeout: typeof timers.setTimeout =
+  globalThis.setTimeout || setTimeoutFallback;
+export const setInterval: typeof timers.setInterval =
+  globalThis.setInterval || setIntervalFallback;
+
+export const active = notImplemented("timers.active");
+export const _unrefActive = notImplemented("timers._unrefActive");
+export const enroll = notImplemented("timers.enroll");
+export const unenroll = notImplemented("timers.unenroll");
```
pi0, line 37: For exports that are identical to unenv built-in, I think we can simply reexport.

src/runtime/node/timers/internal/set-immediate.ts
pi0, line 1: That's fine for PR but please try to reduce diff in PRs if it is only code style change.

vicb:
Thanks for the review and help @IgorMinar and @pi0.

I have updated the PR according to the feddback.

vicb:
> bah! I'm sorry @vicb, this must have been frustrating to debug. I'm glad that you got to the bottom of it and resolved it. Thank you!

It was fun to scrath our heads with @petebacondarwin this morning 🤔

src/runtime/node/timers/internal/set-interval.ts
```diff
@@ -1,18 +1,17 @@
...
+  value?: T,
+): AsyncIterable<T> {
+  yield value as T;
+}
 
 export function setIntervalFallback(
   callback: (args: void) => void,
   ms?: number,
 ): NodeJS.Timeout;
+
```
vicb, line 14: actually this should be better reverted. Let me force push one last time

pi0 (approved):

PR #303, opened 2024-09-06
feat(node:timer): implement `setImmediate` with `setTimeout`

vicb:
As discussed in https://github.com/unjs/unenv/pull/302 using a timeout is a more sensible behavior

### DIFF
--- a/src/runtime/node/timers/internal/immediate.ts
+++ b/src/runtime/node/timers/internal/immediate.ts
@@ -1,19 +1,32 @@
 export class Immediate<TArgs extends any[]> implements NodeJS.Immediate {
   _onImmediate: (...args: TArgs) => void;
+  private _timeout?: NodeJS.Timeout;
 
   constructor(callback: (...args: TArgs) => void, args: TArgs) {
     this._onImmediate = callback;
-    callback(...args);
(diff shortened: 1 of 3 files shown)


pi0 (approved):
❤️

src/runtime/node/timers/internal/immediate.ts
```diff
@@ -1,19 +1,26 @@
 export class Immediate<TArgs extends any[]> implements NodeJS.Immediate {
   _onImmediate: (...args: TArgs) => void;
+  private _timeout: NodeJS.Timeout;
 
   constructor(callback: (...args: TArgs) => void, args: TArgs) {
     this._onImmediate = callback;
-    callback(...args);
+    this._timeout = setTimeout(callback, 0, ...args);
```
pi0, line 7: One additional check  we can also make sure `globalThis.setTimeout` is available and if not, use previous method.

vicb:
@pi0 Thanks for the help and review. I have updated the PR.

vicb:
and even fixed the ci error... I think ;)

vicb:
@pi0 Thanks for your help and being so reactive 🙏

IgorMinar:
Thanks for getting this fixed!

PR #309, opened 2024-09-16
fix: ensure Cloudflare webcrypto.getRandomBytes is bound to the correct this object

petebacondarwin:
It appears that this is the only function in the webcrypto exports list that
needs to be bound to its original object.

Note that it must be bound to the `webcrypto` object even though workerd also
exports it on the `crypto` namespace directly.

### DIFF
--- a/src/runtime/node/crypto/$cloudflare.ts
+++ b/src/runtime/node/crypto/$cloudflare.ts
@@ -89,7 +89,6 @@ export const {
   getDiffieHellman,
   getFips,
   getHashes,
-  getRandomValues,
   hkdf,
   hkdfSync,
   pbkdf2,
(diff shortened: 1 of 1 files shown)


src/runtime/node/crypto/$cloudflare.ts
```diff
@@ -106,18 +105,21 @@ export const {
   timingSafeEqual,
 } = workerdCrypto;
 
+// Special case getRandomValues as it must be bound to the webcrypto object
+export const getRandomValues = workerdCrypto.getRandomValues.bind(
+  workerdCrypto.webcrypto,
+);
```
petebacondarwin, line 111: This is the correct fix.
We need to ensure that every export of this function is correctly bound.
Not just the one from webcrypto.

pi0 (approved):

PR #316, opened 2024-10-07
feat(node/net): implement `Server` mock

### LINKED ISSUE #314: Ambigious error messages cause false assumptions on EventEmitter.
### Environment

Cloudflare 

### Reproduction

Create a TLS server and call `.once()` function

### Describe the bug

Unimplemented NET server does not extend EventEmitter.

### Additional context

Using TLS server `.once()` function will throw an error hinting towards, `socket.once` does  not exist. Some folks are thinking that Cloudflare workers does not support node:events and particularly `once` function because of this error. It indeed supports `once()` and `emitter.once()`

Please look into https://alexbevi.com/blog/2024/09/11/why-cloudflare-workers-dont-work-with-mongodb/

### Logs

_No response_

pi0:
resolves #314 (aka PI issue!)

`net.Server` (and `tls.Server` extending it) mock is implemented and extending EventEmitter to have more explicit behavior.

### DIFF
--- a/src/runtime/node/net/internal/server.ts
+++ b/src/runtime/node/net/internal/server.ts
@@ -1,0 +1,45 @@
+import type * as net from "node:net";
+import { EventEmitter } from "node:events";
+
+// Docs: https://nodejs.org/api/net.html#net_class_net_socket
+export class Server extends EventEmitter implements net.Server {
+  readonly __unenv__ = true;
+
+  maxConnections: number = 1;
+  connections: number = 0;
+  readonly listening: boolean = false;
+
+  constructor(
+    arg1?: net.ServerOpts | ((socket: net.Socket) => void),
+    arg2?: (socket: net.Socket) => void,
+  ) {
+    super();
+  }
+
+  listen(): this {
+    return this;
+  }
+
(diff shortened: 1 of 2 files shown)


src/runtime/node/net/internal/server.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  constructor(
+    arg1?: net.ServerOpts | ((socket: net.Socket) => void),
+    arg2?: (socket: net.Socket) => void,
+  ) {
+    super();
+  }
+
+  listen(): this {
+    return this;
```
anonrig, line 20: Maybe we should just call notImplemented here and throw error?

pi0: Added fail: d364dec83664f56da994f275f90c02f40b5a6ed8

My main intention was to not fail if any lib is using listen for exmaplte to test a random port. But being explicit is probably bettter.

anonrig: If this doesn't throw an error, probably their own code will fail at some point, and they'd assume something similar to the original issue that this PR fixes.

anonrig (approved):

PR #317, opened 2024-10-07
test: bump compatibilityDate past 2024-09-23

vicb:
`unenv` is used by wrangler only when the `workerd` flag `nodejs_compat_v2` is active.

And `nodejs_compat_v2` is implied by `nodejs_compat` after [2024-09-23](https://developers.cloudflare.com/workers/runtime-apis/nodejs/#get-started).

This PR bumps the `compatibilityDate` past 2024-09-23.

`2024-10-04` is the max `compatibilityDate` supported by the packaged `workerd` (=it's build date).

### DIFF
--- a/test/workerd/config.capnp
+++ b/test/workerd/config.capnp
@@ -10,7 +10,7 @@ const testsWorker :Workerd.Worker = (
   modules = [
     (name = "tests", esModule = embed "./tests.mjs")
   ],
-  compatibilityDate = "2024-09-01",
+  compatibilityDate = "2024-10-04",
   compatibilityFlags = ["nodejs_compat"],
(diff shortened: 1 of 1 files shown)


pi0:
Thanks. BTW the tests purely use workerd

vicb:
Thank you!

> BTW the tests purely use workerd

I get that.
`nodejs_compat_v2` is a workerd flag affecting [its behavior](https://github.com/cloudflare/workerd/blob/77ebaf02be55b9867cb2c65aee16fc6f183bb411/src/workerd/io/compatibility-date.capnp#L489C1-L495C77) and having this active will match the behavior wrangler user see.

PR #319, opened 2024-10-09
fix(cloudflare): disable hybrid `perf_hooks`

vicb:
…oks` (#257)"

It's causing trouble in wrangler, see https://github.com/cloudflare/workers-sdk/pull/6920#issuecomment-2400451231


src/presets/cloudflare.ts
```diff
@@ -69,17 +68,6 @@ const cloudflarePreset: Preset = {
...
-    Performance: ["perf_hooks", "Performance"],
-    PerformanceEntry: ["perf_hooks", "PerformanceEntry"],
-    PerformanceMark: ["perf_hooks", "PerformanceMark"],
-    PerformanceMeasure: ["perf_hooks", "PerformanceMeasure"],
-    PerformanceObserver: ["perf_hooks", "PerformanceObserver"],
-    PerformanceObserverEntryList: [
-      "perf_hooks",
-      "PerformanceObserverEntryList",
-    ],
-    PerformanceResourceTiming: ["perf_hooks", "PerformanceResourceTiming"],
```
pi0, line 82: Since the cloudflare variant is only broken, I think disabling these would be enough (and hybrid from L32) should be enough?

Happy to reiterate in another PR if it is not

pi0 (approved):
Thanks.

PR #311, opened 2024-09-19
test: setup workerd isolate tests

pi0:
This PR adds isolated tests via workerd.

This setup uses a small Node.js wrapper that starts a module fallback server (based on https://github.com/cloudflare/workerd/pull/1423) and then spawns `workerd test` to run tests.

The fallback server, maps `unenv/runtime/*` requests to source code (transpiled with esbuild to allow HMR). esbuild config uses `env(nodeless, cloudflare)` aliases to match the expected bundler behavior when resolving internal `node:` aliases.  

As sample, two tests for `crypto.getRandomValues` and `url.parse` (fail behavior) are added. We shall iterate and add more coverage for runtime safety as we introduce new features.

Usage:

- `pnpm test:workerd`
- `pnpm test:workerd --watch` (watch mode)
- `DUMP_MODULES=1 pnpm test:workerd` (dump server modules to `test/workerd/.tmp` for inspection)

PR #321, opened 2024-10-09
refactor(node/buffer): update cloudflare hybrid exports

anonrig:
Let's reflect the truth for cloudflare buffer hybrid module. Please validate before landing this.

### DIFF
--- a/src/runtime/node/buffer/$cloudflare.ts
+++ b/src/runtime/node/buffer/$cloudflare.ts
@@ -1,15 +1,9 @@
 // https://nodejs.org/api/buffer.html
 import type nodeBuffer from "node:buffer";
 
-export {
-  INSPECT_MAX_BYTES,
-  resolveObjectURL,
-} from "./index";
+export { INSPECT_MAX_BYTES, resolveObjectURL } from "./index";
 
-import {
-  INSPECT_MAX_BYTES,
-  resolveObjectURL,
-} from "./index";
+import { INSPECT_MAX_BYTES, resolveObjectURL } from "./index";
 
 // @ts-ignore typings are not up to date, but this API exists, see: https://github.com/cloudflare/workerd/pull/2147
 const workerdBuffer = process.getBuiltinModule("node:buffer");
@@ -26,7 +20,7 @@ export const {
   isUtf8,
   kMaxLength,
   kStringMaxLength,
-  transcode
+  transcode,
 } = workerdBuffer;
 
 export default {
--- a/src/runtime/node/buffer/$cloudflare.ts
+++ b/src/runtime/node/buffer/$cloudflare.ts
@@ -2,27 +2,13 @@
 import type nodeBuffer from "node:buffer";
 
 export {
-  Blob,
-  File,
   INSPECT_MAX_BYTES,
-  atob,
-  btoa,
-  isAscii,
-  isUtf8,
   resolveObjectURL,
-  transcode,
 } from "./index";
 
 import {
-  Blob,
-  File,
   INSPECT_MAX_BYTES,
-  atob,
-  btoa,
-  isAscii,
-  isUtf8,
   resolveObjectURL,
-  transcode,
 } from "./index";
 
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for PR. But with this, if there is a code that imports `node:buffer`, it will loose access to the same exports from Node.js.

```
$ Object.keys(require('buffer'))
[
  'Buffer',            'SlowBuffer',
  'transcode',         'isUtf8',
  'isAscii',           'kMaxLength',
  'kStringMaxLength',  'btoa',
  'atob',              'constants',
  'INSPECT_MAX_BYTES', 'Blob',
  'resolveObjectURL',  'File'
]
```

Can you please clarify the goal of this?

anonrig:
I'm trying to get myself familiar with `unenv` repository :) I've updated the pull-request.

src/runtime/node/buffer/$cloudflare.ts
```diff
@@ -2,35 +2,21 @@
 import type nodeBuffer from "node:buffer";
 
 export {
-  Blob,
-  File,
   INSPECT_MAX_BYTES,
-  atob,
-  btoa,
```
petebacondarwin, line 9: Why do we no longer export these (and others) when they are neither provided by workerd?

anonrig: They are provided by workerd: https://github.com/cloudflare/workerd/blob/main/src/node/buffer.ts

src/runtime/node/buffer/$cloudflare.ts
```diff
@@ -1,37 +1,27 @@
...
 
 // @ts-ignore typings are not up to date, but this API exists, see: https://github.com/cloudflare/workerd/pull/2147
 const workerdBuffer = process.getBuiltinModule("node:buffer");
 
 // TODO: Ideally this list is not hardcoded but instead is generated when the preset is being generated in the `env()` call
 //       This generation should use information from https://github.com/cloudflare/workerd/issues/2097
-export const { Buffer, SlowBuffer, constants, kMaxLength, kStringMaxLength } =
-  workerdBuffer;
+export const {
+  Buffer,
```
vicb, line 14: please sort lines - use a VSCode plugin ;)

vicb (changes requested):

src/runtime/node/buffer/$cloudflare.ts
anonrig: Unfortunatelly, I don't use VSCode, but I'll manually update it :-))

src/runtime/node/buffer/$cloudflare.ts
vicb: I had the same comment... hard to keep up with your simultaneous updates 🤣
Maybe you can add `fixup!` commits instead of fore pushing.

anonrig: 060ec21017e667e3f8041664f8fb7b6b9cc8bbaee97d2e4d120a4871b50911a5

pi0 (approved):
❤️

src/runtime/node/buffer/$cloudflare.ts
pi0: We squash and also have a nice fellow ([autofix.ci](https://autofix.ci/)) that pushes fixes.

vicb (changes requested):
The updates looks great.

Thanks to @pi0, it's now easy to add tests (see #311).

Could you please add a test (i.e. asserting the types of the moved exports).

anonrig:
> Could you please add a test (i.e. asserting the types of the moved exports).

Any suggestions on where and how to test them?

vicb:
> Any suggestions on where and how to test them?

My suggestion was to check #311 -> https://github.com/unjs/unenv/blob/main/test/workerd/tests.mjs

anonrig:
@vicb Would you mind re-reviewing this?

PR #325, opened 2024-10-11
fix(node/buffer): add coudflare exports for `atob` and `btoa` and add tests

anonrig:
Fixes tests for cloudflare buffer. Reverts only `atob` and `btoa` changes.

Ref: https://github.com/unjs/unenv/pull/324

PR #326, opened 2024-10-11
chore: add cloudflare team to workerd tests codeowners

IgorMinar:
This came out of https://github.com/unjs/unenv/pull/325 which we should be able to approve and merge without @pi0  but we can't right due to the OWNERS config not being up todate.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,3 +2,4 @@
 
 # all cloudflare files including src/preset/cloudflare.ts, **/$cloudflare.ts, etc.
 **/*cloudflare** @unjs/cloudflare
+/test/workerd/** @unjs/cloudflare


anonrig (approved):

pi0:
I wish this were discussed in one of the proper channels we already have but here is a quick context:

Nitro (and Nuxt) use unenv injections by default even if users don't specify `node_compat` in their wrangler config (which is opt-in) we need to make sure tests also cover for this situation.

In the next steps, I am planning to enable the same tests in a matrix with/without `node_compat` because of this.

I think it is fair to ask until we do this. So that afterwards when the cloudflare team updates tests making sure they also cover nodeless state.

vicb (approved):

pi0 (approved):
Approving temporarily until we find a better solution. 

@unjs/cloudflare **please** help to make sure tests cover both nodeless and cloudflared situation.

Issue #262, opened 2024-06-18
Dedicated `wrangler` preset


pi0:
We have introduced a new `cloudflare` preset and set of `$cloudflare` polyfill variations set by this preset to allow leveraging hybrid Node.js compatibility available in workerd runtime.

Initially, `cloudflare` preset was introduced to be included in Nitro (rollup based) and we are working with cloudflare team to also integrate unenv into wrangler tooling (esbuild based)

Apart from bundler subtle differences (rollup and esbuild), from discussions like [#257](https://github.com/unjs/unenv/pull/257) i can predict in the future, we might have legit reasons that Nitro and Wrangler might need different ways of adding Cloudflare polyfills.

One way is to expose standard plugin (~> https://github.com/unjs/unenv/issues/263) but still decisions might vary on tool level between wrangler and Nitro objectives.

I propose to introduce a new `wrangler` preset ahead of time (maintained by @IgorMinar @jculvey as well) to allow forward-compatible way to collaborate releases between unenv and wrangler without imposing unwanted changes to the end users.

IgorMinar:
I'll think about this more, but I'd really like to avoid wrangler specific preset. The hybrid polyfill we implemented relies on aligning versions of workerd and unenv and this will become much harder to reason about if we in addition to this also have to think about two different presets.

pi0:
I also wish minimum possible variations and `wrangler` preset (extending `cloudflare > nodeless`) to be as empty as possible.

My intention is not to duplicate efforts but have a safe space that we can put wrangler-specific requirements. For example injection for global `Performance` polyfills with Node.js addons. 

(i might have better idea for configurable presets to avoid additional one but the gist of idea will be same: a way to clearly know if we are exporting preset for wrangler or nitro and with or without Node global API)

Dmytro-Tihunov:
hey @pi0 looking for solution for 4 days, using Nuxt and cloudflare deployment and get this "[unenv] crypto.createHash is not implemented yet!" what shall i do ?

pi0:
@Dmytro-Tihunov please report issue in either Nuxt or Nitro repository with reproduction and avoid pining in other threads 🙏🏼

pi0:
Cloudflare / Wrangler preset is being externalized #336

PR #329, opened 2024-10-11
refactor(cloudflare): make `node:util/types` a `nodejs_compat` module

anonrig:
All methods in `util/types` are implemented. We can remove the hybrid module status from `node:util/types`


jasnell (approved):

src/presets/cloudflare.ts
```diff
@@ -19,6 +19,7 @@ const cloudflareNodeCompatModules = [
   "stream/promises",
   "stream/web",
   "string_decoder",
+  "util/types",
```
pi0, line 22: Just checking, do you have any upstream test that makes sure `util/types` will support future additions from Node?

anonrig: Workerd doesn't run Node.js in the test suite. The missing functionality will be tracked with @jasnell and mine best efforts. FYI: it's highly unlikely to have any new methods in util/types.

test/workerd/tests.mjs
```diff
@@ -58,3 +58,13 @@ export const buffer_implements = {
     assert.ok(new Buffer.Blob([]));
   },
 };
+
+// --- node:util
+
+export const util_implements = {
+  async test() {
+    const { types } = await import("unenv/runtime/node/util");
+    assert.strictEqual(types.isExternal("hello world"), false);
```
pi0, line 67: Is there a simple test that we can add that the return value is `true` today?

I checked cloudflare/workerd#2264 but no clues and [node docs](https://nodejs.org/api/util.html#utiltypesisexternalvalue) uses node bindings as example. does workerd support napi?

I am asking this mainly because if there are really no real cases, i was thinking the nodeless polyfill (when node_compat is not enabled) could be `false` perhaps?

anonrig: there is no real case. workerd doesn't support napi.

pi0: then is it safe to assume it always returns `false`?

anonrig: I think so. @jasnell wdyt?

pi0: If in case we agree (i think you are experts to say this of course!), i would suggest to update [here](https://github.com/unjs/unenv/blob/main/src/runtime/node/util/internal/types.ts#L4) to reflect this and always return `false`.

This way when we enable matrix for nodeless, tests pass.

vicb:
What workerd version this PR depends on?

We can not pull this code in wrangler before wrangler pulls the required version of workerd.

More generally I think we should use whatever workerd version is bundled with wrangler for the tests. Having a different version is a recipe for troubles.

@anonrig that's probably something we need to discuss internally.

pi0:
I think we discussed about it to split wrangler preset/versioning so we can test and upgrade unenv according to the latest workerd release.

vicb:
> I think we discussed about it to split wrangler preset/versioning so we can test and upgrade unenv according to the latest workerd release.

Are you referring to #262?

I think the problem we need to solve for wrangler is how to make sure `unenv`  is compatible/tested with the version version of workerd it uses. 

I'm unsure if creating a wrangler preset is the best way to solve this - hopefully this repo is never aware about wrangler. What I mean is that unenv patches should depend on workerd version, not wrangler version. When we have a way to do that then wrangler will pick the unenv version matching the workerd version it bundles.

test/workerd/tests.mjs
jasnell: workerd currently does not expose any `External`s to public API that can be tested.

anonrig:
> More generally I think we should use whatever workerd version is bundled with wrangler for the tests. Having a different version is a recipe for troubles.

@vicb I agree. In the long term, we should match those 2 version. Right now, wrangler `main` and unenv `main` has different workerd version.

vicb:
> > More generally I think we should use whatever workerd version is bundled with wrangler for the tests. Having a different version is a recipe for troubles.
> 
> @vicb I agree. In the long term, we should match those 2 version. Right now, wrangler `main` and unenv `main` has different workerd version.

Maybe we should have 2 tags for workerd: "prod" and "wrangler" and test both versions.

pi0 (approved):
Thank you ❤️ 

Feel free to merge when/if prod version of workerd shipped this

PR #334, opened 2024-10-14
refactor(node/buffer): update buffer to be cloudflare compat module

anonrig:
Latest version of workerd adds all missing functionality of `node:buffer` and exposes `node:sys` and `sys` modules.

FYI: This requires workerd [v1.20241011.0](https://github.com/cloudflare/workerd/releases/tag/v1.20241011.0), which is included in the main branch of unenv, but not in the main branch of wrangler.


vicb:
> FYI: This requires the latest version of workerd, which is included in the main branch of unenv, but not in the main branch of wrangler.

Could you specify the exact version and mark this as draft until wrangler is updated?

test/workerd/tests.mjs
```diff
@@ -36,14 +36,12 @@ export const url_parse = {
 export const buffer_implements = {
```
vicb, line 36: I don't think this test in unenv makes any sense any more?

anonrig:
> Could you specify the exact version and mark this as draft until wrangler is updated?

These changes are included in https://github.com/cloudflare/workerd/releases/tag/v1.20241011.0.

test/workerd/tests.mjs
```diff
@@ -60,5 +58,18 @@ export const buffer_implements = {
...
     assert.ok(new buffer.File([], "file"));
     assert.ok(new buffer.Blob([]));
+    assert.strictEqual(typeof buffer.INSPECT_MAX_BYTES, "number");
+    assert.strictEqual(typeof buffer.resolveObjectURL, "function");
+  },
+};
+
+// --- node:sys
+
+export const sys_implements = {
```
vicb, line 68: Same here.

Actually maybe the only thing we need to test for both modules is that

```
await import("unenv/runtime/node/buffer") === createRequire('/')('node:buffer');
```

test/workerd/tests.mjs
anonrig: It's not needed but I'm also not sure if removing it is the best course of action, since this tests that the CF workers has the following functions.

test/workerd/tests.mjs
anonrig: I don't follow. Can you recommend a `code suggestion` on the github UI?

test/workerd/tests.mjs
vicb: > since this tests that the CF workers has the following functions.

It tests nothing else than workerd so I think test should live in workerd (where everything is probably already tested).

I fail to see what the current would ever catch that workerd test suite would not?

anonrig: That's true. I'll remove it.

test/workerd/tests.mjs
vicb: The idea here is that `require("node:buffer)` would be processed by unenv while `createRequire('/')('node:buffer')` would return the actual workerd module.

So it would test that unenv does not change the workerd buffer module.

anonrig: Ah, makes sense. Thank you!

anonrig:
cc @unjs/cloudflare

test/workerd/tests.mjs
```diff
@@ -35,31 +36,21 @@ export const url_parse = {
...
+  async test() {
+    const sys = await import("node:sys");
+    assert.ok(sys != null);
+    const plainSys = await import("sys");
+    assert.ok(plainSys != null);
```
vicb, line 53:
```suggestion
    assert.strictEqual(
      await import("node:sys"),
      createRequire("/")("node:sys"),
    );
```

Can't we do the same as for buffer here?

test/workerd/tests.mjs
```diff
@@ -35,31 +36,21 @@ export const url_parse = {
...
-      assert.strictEqual(
-        dest.toString(),
-        Buffer.from("těst ☕", "utf8").toString(),
-      );
-    }
-    assert.ok(new buffer.File([], "file"));
-    assert.ok(new buffer.Blob([]));
+    assert.strictEqual(
+      await import("node:buffer"),
+      createRequire("/")("node:buffer"),
```
pi0, line 41: Is the purpose of this change mainly to check `$cloudflare` entry is the same as the workerd impl?

If yes, please let's keep the actual buffer tests. I have added workerd tests not only to test cloudflare hybrid (with node_compat opt-in) but also to test polyfills in an environment that is not Node.js.

This tests in this file have both purposes.

anonrig:
@pi0 Everytime we change the test file, it seems your review is required in order to land it. Can you review and land this pull-request please?

test/workerd/tests.mjs
vicb: There is no more `$cloudflare`

vicb: > If yes, please let's keep the actual buffer tests. I have added workerd tests not only to test cloudflare hybrid (with node_compat opt-in) but also to test polyfills in an environment that is not Node.js.

We might keep forgetting that.
What about creating 2 distincts sub-folders?

vicb:
> @pi0 Everytime we change the test file, it seems your review is required in order to land it. Can you review and land this pull-request please?

See https://github.com/unjs/unenv/pull/326#issuecomment-2408149169

test/workerd/tests.mjs
vicb: It's great to add the reason why the convo is resolved for history.

pi0:
> Everytime we change the test file, it seems your review is required in order to land it. Can you review and land this pull-request please?

Dear @anonrig i have already explained it in https://github.com/unjs/unenv/pull/326#issuecomment-2408149169 Workerd tests added in #311 are both for wrangler/node_compat and unenv (polyfill) isolate tests itself. unenv is a shared project. No one did a review on tests PR, at least you could reach out to me asking it's purpose before trying to push changes on it and saying it is a blocker.

Sadly today only because of supporting wrangler SDK, unenv got completely broken for it's main users (Nuxt and Nitro). i hear no single response from anyone of cloudflare team about it for weeks yet i am at airport, offday, evening, doing prompt reviewers to help not making situation worst than this.

Sorry for telling all these but i am trying to do all my best to unblock your team while wrangler, blocked unenv project to be used for it's primary user (Nuxt and Nitro), i hope it can help you understand situation.

test/workerd/tests.mjs
pi0: Ideally, we should have a matrix (run same tests in two conditions). 

But TBH more I'm thinking we really needed separated versioning. (working on a draft right now to unblock you)

southpolesteve:
@pi0 👋 One of the Workers EMs here. First off, I know you're doing a ton of work here to help unblock us. Its much appreciated. Fully acknowledge this is a shared project with many users not just wrangler. 

Sorry for the lack of communication on our side. I think we've got couple of wires crossed internally and I'll work with the team to resolve that. 

You mentioned in the other thread "proper channels". Can you add me to wherever that is? I am happy to jump in to make sure talking more and doing everything we can do support the shared needs of the project. `sfaulkner@cloudflare.com` is my email

pi0:
Hi, dear Steve @southpolesteve I appreciate your help on this matter.

We mainly used discord channels to communicate and i mentioned our blockers many times both public and private. I am more than happy to find any better way to improve our communication, we all love cloudflare! If you happen to be on discord, i believe someone from the team can send you an invitation (or please direct me `@pi0`).

In the meantime, let's unblock this situation first.

pi0:
Temporarily merged #326 until finding a better solution.

@anonrig if you can kindly help to maintain tests coverage for nodeless that would be best otherwise please feel free to merge it as hotfix and iterate.

test/workerd/tests.mjs
```diff
@@ -36,14 +36,12 @@ export const url_parse = {
 export const buffer_implements = {
   async test() {
     const encoder = new TextEncoder();
-    const buffer = await import("unenv/runtime/node/buffer");
+    const buffer = await import("node:buffer");
```
anonrig, line 39: @pi0 How do you propose covering `nodeless` support for this change? Should this line be reverted?

pi0: Maybe having two test exports in the meantime?

(if too hard please keep the version you prefer for now)

anonrig: I can update the PR.

anonrig:
I've updated the tests to have 2 different paths. @pi0

src/runtime/node/buffer/index.ts
```diff
@@ -22,8 +22,8 @@ export const transcode = notImplemented("buffer.transcode");
 export const isUtf8 = notImplemented("buffer.isUtf8");
 export const isAscii = notImplemented("buffer.isAscii");
 
-export const btoa = global.btoa;
-export const atob = globalThis.atob;
+export const btoa = global.btoa.bind(global);
```
vicb, line 25: For consistency:

```suggestion
export const btoa = globalThis.btoa.bind(globalThis);
```

But because this is a fix it should better be in a different PR (together with the test).

test/workerd/tests.mjs
```diff
@@ -33,17 +33,32 @@ export const url_parse = {
...
+    assert.strictEqual(typeof buffer.transcode, "function");
+    assert.strictEqual(typeof buffer.File, "function");
+    assert.strictEqual(typeof buffer.Blob, "function");
+    assert.strictEqual(typeof buffer.INSPECT_MAX_BYTES, "number");
+    assert.strictEqual(typeof buffer.resolveObjectURL, "function");
+    assert.strictEqual(typeof Buffer.from, "function");
+  },
+};
+
+export const workerd_implements_buffer = {
```
vicb, line 53: Not sure if this test has any value as this is already tested in workerd - the infra only runs this code on workerd.

test/workerd/tests.mjs
```diff
@@ -60,6 +75,19 @@ export const buffer_implements = {
...
     assert.ok(new buffer.File([], "file"));
     assert.ok(new buffer.Blob([]));
+    assert.strictEqual(typeof buffer.INSPECT_MAX_BYTES, "number");
+    assert.strictEqual(typeof buffer.resolveObjectURL, "function");
+  },
+};
+
+// --- node:sys
+
+export const sys_implements = {
```
vicb, line 85: Maybe you can collapse the previous test in this one and rename it `workerd_modules`, only checking that the import is not null (or would it throw?)

src/runtime/node/buffer/index.ts
anonrig: One of the tests are failing because workerd throws an error for `invalid this`. I'll apply your suggestion. Thanks for the review.

test/workerd/tests.mjs
anonrig: I've kept it due to @pi0's request/suggestion.

jasnell (approved):

src/presets/cloudflare.ts
```diff
@@ -19,6 +20,7 @@ const cloudflareNodeCompatModules = [
   "stream/promises",
   "stream/web",
   "string_decoder",
+  "sys",
```
vicb, line 23: can we revert this together with l 50-52?

- the PR is "update buffer to be cloudflare compat module", let's not mixed things
- if `sys` is an alias of `util`, why do we want `util` to be a node compat module while `sys` is hybrid

vicb (approved):
Thanks for the PR Yagiz.

Could we remove the `sys` change from here?

PR #338, opened 2024-10-17
fix(buffer): bind btoa and atob to globalThis

vicb:
This PR adds a change that was initially in @anonrig  #334 but we had to extract as we do not own the code.


anonrig (approved):

pi0 (approved):

pi0:
Thanks!

PR #342, opened 2024-10-23
refactor(test/workerd): remove dead code

vicb:


pi0:
Thanks for PR. I have mainly kept them as future reference to know what params are available since planning to generalize this script in the future as a pkg.

I think we can keep them all in a typed context so they are not dead code.

PR #343, opened 2024-10-26
feat(cloudflare): remove unnecessary aliases

anonrig:
`assert/strict` and `sys` aliases are available on workerd. we don't need unenv to handle them.

### DIFF
--- a/src/presets/cloudflare.ts
+++ b/src/presets/cloudflare.ts
@@ -11,10 +11,13 @@ const cloudflareNodeCompatModules = [
   "_stream_transform",
   "_stream_writable",
   "assert",
+  "assert/strict",
   "buffer",
   "diagnostics_channel",
   "events",
   "path",
+  "path/posix",
+  "path/win32",
   "querystring",
   "stream",
   "stream/consumers",
@@ -23,6 +26,7 @@ const cloudflareNodeCompatModules = [
   "string_decoder",
   "url",
   "util/types",
+  "sys",
   "zlib",
 ];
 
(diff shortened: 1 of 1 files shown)


jasnell (approved):

src/presets/cloudflare.ts
```diff
@@ -46,12 +50,6 @@ const cloudflarePreset: Preset = {
         [`node:${p}`, `node:${p}`],
       ]),
     ),
-    // The `node:assert` implementation of workerd uses strict semantics by default
-    "assert/strict": "node:assert",
-    "node:assert/strict": "node:assert",
-    // The `node:sys` module is just a deprecated alias for `node:util` which we implemented using a hybrid polyfill
-    sys: "unenv/runtime/node/util/$cloudflare",
```
vicb, line 53: @anonrig I think we already discussed that one and the conclusion was that we need it because not all functions are implemented. It should not be a node compat module as long as util is not. Could you please revert this?

src/presets/cloudflare.ts
```diff
@@ -11,10 +11,13 @@ const cloudflareNodeCompatModules = [
   "_stream_transform",
   "_stream_writable",
   "assert",
+  "assert/strict",
   "buffer",
   "diagnostics_channel",
   "events",
   "path",
+  "path/posix",
```
vicb, line 19: @anonrig has there been any changes on the worked side related to that? I think I tried exactly before messaging you earlier this week and that was not working - could you add a test for i.e. separator

PR #344, opened 2024-10-27
fix(cloudflare): revert sys alias change

anonrig:
Partially reverts https://github.com/unjs/unenv/pull/343

### DIFF
--- a/src/presets/cloudflare.ts
+++ b/src/presets/cloudflare.ts
@@ -26,7 +26,6 @@ const cloudflareNodeCompatModules = [
   "string_decoder",
   "url",
   "util/types",
-  "sys",
   "zlib",
 ];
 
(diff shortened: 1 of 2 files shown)


src/presets/cloudflare.ts
```diff
@@ -51,6 +51,11 @@ const cloudflarePreset: Preset = {
       ]),
     ),
 
+    // The `node:sys` module is just a deprecated alias for `node:util`
```
vicb, line 54: Please remove l29 as well. 
Thanks

test/workerd/tests.mjs
```diff
@@ -102,3 +102,18 @@ export const util_implements = {
     assert.strictEqual(types.isAnyArrayBuffer(new ArrayBuffer(0)), true);
   },
 };
+
+// --- node:path
+
+export const path_implements = {
```
vicb, line 108: I think was I tested and was not working was path and directory separators

vicb (changes requested):
Thanks for the updates, see my inline comments

test/workerd/tests.mjs
anonrig: @jasnell's pull-request addresses this https://github.com/cloudflare/workerd/pull/3015

vicb: Changes required:
- pathModule.~default~
- Update `workerd`

-> we won't be able to merge this until workerd is updated. Could you please split this PR in 2 different PRs so that we can revert the `sys` change ASAP.

PR #345, opened 2024-10-28
fix(cloudflare): Use polyfills for path/*

vicb:
Partial revert of #343 until we have a proper fix in workerd, see https://github.com/cloudflare/workerd/issues/3012

/cc @anonrig

### DIFF
--- a/test/workerd/tests.mjs
+++ b/test/workerd/tests.mjs
@@ -102,3 +102,20 @@ export const util_implements = {
     assert.strictEqual(types.isAnyArrayBuffer(new ArrayBuffer(0)), true);
   },
 };
+
+// --- node:path
+
+export const unenv_polyfills_path = {
+  async test() {
+    const pathWin32 = await import("unenv/runtime/node/path/win32");
+    assert.strictEqual(typeof pathWin32.resolve, "function");
+    // Note: unenv uses `unjs/pathe` which behavior differs from Node.js
+    // See https://github.com/unjs/pathe
+    assert.strictEqual(pathWin32.sep, "/");
+    assert.strictEqual(pathWin32.delimiter, ":");
+    const pathPosix = await import("unenv/runtime/node/path/posix");
+    assert.strictEqual(typeof pathPosix.resolve, "function");
+    assert.strictEqual(pathPosix.sep, "/");
+    assert.strictEqual(pathPosix.delimiter, ":");
+  },
+};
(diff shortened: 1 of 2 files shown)


test/workerd/tests.mjs
```diff
@@ -102,3 +102,20 @@ export const util_implements = {
...
   },
 };
+
+// --- node:path
+
+export const unenv_polyfills_path = {
+  async test() {
+    const pathWin32 = await import("unenv/runtime/node/path/win32");
+    assert.strictEqual(typeof pathWin32.resolve, "function");
+    // Note: unenv uses `unjs/pathe` which behavior differs from Node.js
```
petebacondarwin, line 112: If both `path/win32` and `path/posix` work the same way as `path` then why use unenv to polyfill this? Instead just redirect `path/win32` and `path/posix` to the built in `path`.

vicb: 2 things:
- `path/win32` and `path/posix` are provided by `workerd` but there is a bug as they only export `default`. This bug must be fixed
- `pathe` behavior is not consistent with Node so I don't think we want this.

> just redirect `path/win32` and `path/posix` to the built in `path`

That could be a workaround for now until `workerd` is fixed but I think we can live with the current behavior (i.e. before #343) until workerd is fixed.

Ref: https://nodejs.org/api/path.html#windows-vs-posix

vicb: What I mean by "with the current behavior" is that #343 has not been merged into wrangler for now.

The goal of this PR is to keep the exact same behavior for wrangler until we have a proper fix in workerd and then switch to using that.
After workerd is fixed, unenv updated to use workerd directly instead of the polyfill then we can update wrangler. The behavior will then match Node instead of pathe (that is win32 and posix will behave differently).

petebacondarwin: So this is actually just a partial revert of #343 rather than a solution?
In that case, fine. As long as we then fix up workerd.

vicb: > So this is actually just a partial revert of https://github.com/unjs/unenv/pull/343 rather than a solution?

Exactly

anonrig:
@jasnell's pull-request fixes this: https://github.com/cloudflare/workerd/pull/3015

vicb:
> @jasnell's pull-request fixes this: [cloudflare/workerd#3015](https://github.com/cloudflare/workerd/pull/3015)

Could we merge this before the PR is approved and workerd makes its way to wrangler?

More generally I'm in favor of reverting troublesome PRs ASAP.

vicb:
Thanks for the reviews!

PR #348, opened 2024-11-08
feat(cloudflare): use native path/posix and path/win32

vicb:
path/win32 and path/posix are fixed in workerd [1.20241106.0](https://github.com/cloudflare/workerd/releases/tag/v1.20241106.0#:~:text=Add%20named%20exports%20for%20node%3Apath/win32%20and%20node%3Apath/posix%20by%20%40jasnell) which is integrated in wrangler 3.86

/cc @jasnell @anonrig

### DIFF
--- a/test/workerd/tests.mjs
+++ b/test/workerd/tests.mjs
@@ -1,4 +1,4 @@
-import assert, { strictEqual } from "node:assert";
+import assert from "node:assert";
 import { createRequire } from "node:module";
 import process from "node:process";
 
--- a/test/workerd/tests.mjs
+++ b/test/workerd/tests.mjs
@@ -1,4 +1,4 @@
-import assert from "node:assert";
+import assert, { strictEqual } from "node:assert";
 import { createRequire } from "node:module";
 import process from "node:process";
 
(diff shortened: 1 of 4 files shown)


test/workerd/tests.mjs
```diff
@@ -115,6 +115,15 @@ export const unenv_polyfills_path = {
     assert.strictEqual(pathWin32.delimiter, ":");
     const pathPosix = await import("unenv/runtime/node/path/posix");
     assert.strictEqual(typeof pathPosix.resolve, "function");
+  },
+};
+
+export const workerd_path = {
+  async test() {
+    const pathWin32 = await import("node:path/win32");
+    assert.strictEqual(pathWin32.sep, "\\");
```
pi0, line 124: Just curious, is there even a functional reason to use backslash in workers? (the behavior of [pathe](https://github.com/unjs/pathe) is widely tested with 48M+ downloads and really reliable -- we had barely windows path issues after adopting it)

vicb: The main reason is to behave like Node.
`pathe` choice is more likely reasonable in 99.9999% of the cases

vicb: Also there was a bug in `workerd` causing `.sep` to be undefined `undefined`

pi0: My personal opinion (and i will liklely keep doing it for Nitro even in hybrid) is to use unix paths it is much much safer.

I think workerd is sometimes doing behavior that is different from node (like assert import is always in strict mode) -- but anyway if you are happy to do this path it's okay.

anonrig (approved):

pi0 (approved):

test/workerd/tests.mjs
anonrig: path always resolves into path/unix.

pi0: That's nice then what do you think that sep to be consistent and always `/` as well? (some tools, might use it)

anonrig: node:path/win32 is only there for those who explicitly require it, and it doesn't make sense to change the behavior of that. for node:path/unix, the behavior is still same, because the import is explicitly defined. for `node:path` we resolve unix path always because we'll always run prod on linux.

PR #350, opened 2024-11-23
feat: updates for node v22.9 compatibility

vicb:
Following https://github.com/unjs/unenv/pull/349, I updated the node types upstream in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/71216 - this was released as `@types/node` v22.9.3

The cast is module is no longer needed.


src/runtime/node/module/$cloudflare.ts
```diff
@@ -16,7 +16,10 @@ export {
   _resolveFilename,
   _resolveLookupPaths,
   builtinModules,
+  constants,
```
vicb, line 19: @anonrig I think you have plan to enable compile cache, those methods are not implemented in workerd yet, right?
`enableCompileCache`, `getCompileCacheDir`, and related `constants`

anonrig: It is enabled right now. And yes these methods are not implemented.

anonrig (approved):

src/runtime/node/perf_hooks/internal/performance.ts
```diff
@@ -30,6 +30,7 @@ const nodeTiming = {
   loopStart: 63.262_040_972_709_656,
   loopExit: -1,
   idleTime: 305_360.555_328,
+  uvMetricsInfo: { loopCount: 142, events: 137, eventsWaiting: 97 },
```
pi0, line 33:
```suggestion
  uvMetricsInfo: { loopCount: 0, events: 0, eventsWaiting: 970},
```

(i just noticed, other values should be zeroed as well they seem random snapshot...

vicb: I used a snapshot here because:
```
// grabbed from Node.js v22.3.0 using:
//   performance.nodeTiming
```

But it makes sense to zero out, I'll update all the values

vicb:
Thanks for the reviews and feedback.
I have rebased, fixed a couple minor nits.
PTAL @pi0

pi0 (commented):
Thanks!

vicb:
Thank you @pi0 🙏

PR #352, opened 2024-11-29
node-import

vicb:
We should never import Nodejs packages with relative import as unenv will not get a chance to replace those.

Relates to https://github.com/opennextjs/opennextjs-cloudflare/issues/147: It looks like the issue is caused by the unenv implementation of `Readable#[Symbol.AsyncIterator]`:

```ts
  async *[Symbol.asyncIterator](): NodeJS.AsyncIterator<any> {
    throw createNotImplementedError("Readable.asyncIterator");
  }
```

However this should never be used by workerd that has builtin `stream`.

The second commits add a dummy `IncomingMessage#_read` to make testing easier.

### DIFF
--- a/src/runtime/node/http/internal/request.ts
+++ b/src/runtime/node/http/internal/request.ts
@@ -51,6 +51,8 @@ export class IncomingMessage extends Readable implements http.IncomingMessage {
   get trailersDistinct() {
     return _distinct(this.trailers);
   }
+
+  _read() {}
 }
 
 function _distinct(obj: Record<string, any>) {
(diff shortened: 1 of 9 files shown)


pi0:
This breaks the direct fetch mechanism of nitro, h3, and unenv itself in the Node.js environment ([see this](https://srvx.unjs.io/guide/node#reverse-compatibility) for ref), the reason is that Node.js polyfills of HTTP module depend on in unenv depend on each other.

I think we need to find a different way to solve worked situation and if you can kindly please open an issue with reproduction first I would be happy to help but so sorry this is really not a possible solution.

pi0:
(BTW generally we migrated and should change all other imports to `node:`, I noticed few `node:events` were relative we should fix only http/net/streams are that currently are _functional_ with each other)

Issue #353, opened 2024-11-29
cloudflare: importing `node:http` wrongly import the unenv stream polyfill 


vicb:
### Environment

https://github.com/unjs/unenv/commit/7a9f230e9c74783647ab598796e7da66be854f0f
Nodejs 22.11.0

### Reproduction

See https://github.com/unjs/unenv/pull/352/files#diff-2f5c73584260f31cbdff9f65cf07fd8f619f8b48ddda9d87e1e99d583ac0f87c

### Describe the bug

Without the PR, importing `node:http` also imports the unenv polyfill for stream because it is imported as a relative path.

Because the `Readable` polyfill does not implement the async iterator the test would throw `[unenv] Readable.asyncIterator is not implemented yet!`.

For cloudflare, the builtin stream implementation should be used instead.

### Additional context

See the comments in the [tentative PR](https://github.com/unjs/unenv/pull/352)

vicb:
I am happy to help with this but I would need more info:

> This breaks the direct fetch / http call mechanism of nitro, h3, and unenv itself in the Node.js environment ([see this](https://srvx.unjs.io/guide/node#reverse-compatibility) for ref), the reason is that Node.js polyfills of node:http in unenv depend on other node: polyfills (such as node:net) which are compatible with each other changed in this PR.

I am not sure to understand the problem here.
Is there some code/tests I could run to figure out?

> I think we need to find a different way to solve worked situation (simplest way I would imagine is to add any missing API to polyfills) and if you can kindly please open an issue with reproduction first I would be happy to help but so sorry this is really not a possible solution.

We don't want to use the stream polyfill as it's builtin workerd.
What we want is that the http polyfill pulls in the builtin impl.
One solution could be to have a `$cloudflare-internal` containing the code from the PR - `internal` would be left unchanged and used outside of cloudflare.

> (BTW generally we migrated and should change all other imports to node:, I noticed few node:events were relative we should fix only http/net/streams are that currently are functional with each other)

I changed the relative imports to event in the tentative, are you saying I shouldn't have?
The rationale is the same: worked has a builtin implementation of events and we don't want to include the polyfill transitively.

pi0:
> Is there some code/tests I could run to figure out?

[...]

Issue #323, opened 2024-10-10
https.request for cloudflare


fl0wo:
### Describe the feature

Hi, just to make sure, is[ https.request ](https://github.com/unjs/unenv/blob/main/src/runtime/node/https/index.ts#L17) implemented for cloudflare?

If not, do you think there an ETA for it? I'd love to help if possible.

### Additional information

- [X] Would you be willing to help implement this feature?

fl0wo:
Any update on this?

keiwanmosaddegh:
I'm also interested in this.
@fl0wo have you found any workaround in the meantime?

vicb:
> Hi, just to make sure, is[ https.request ](https://github.com/unjs/unenv/blob/main/src/runtime/node/https/index.ts#L17) implemented for cloudflare?

confirmed that it is not.

Could you please expand on the use case?

vicb:
As you probably already knows Workers use the [workerd Runtime](https://github.com/cloudflare/workerd) that does not implement [all of the Node.js APIs](https://developers.cloudflare.com/workers/runtime-apis/nodejs/).

However we are looking at supporting more and more of the Node libs.

There are 2 things that can help here:
- cloudflare implementing the missing APIs 
- libraries being updated to run on the edge by using Web APIs instead of the Node APIs

So what you can do here to help is:
- Create an issue on the firebase repo explaining that you want to run the code at the edge and asking for Web APIs support (i.e. `fetch`)
- Create [a discussion](https://github.com/cloudflare/workerd/discussions/categories/npm-package-requests) on the workerd repo explaining what repo/version/API you would like to see supported - please link to this issue, the issue on the firebase repo and mention me there

And we'll get that working.

Thanks!

pi0:
Ideallt, i would love to have a polyfill based on [wintercg web sockets proposal](https://sockets-api.proposal.wintercg.org/) which workerd currently support ([startTLS](https://developers.cloudflare.com/workers/runtime-apis/tcp-sockets/#opportunistic-tls-starttls)) — and it should enable a universal / spec compatible solution in the future.

Native workerd support would be cool but i think it is still viable to be discussed for unenv support:)

fl0wo:
> I'm also interested in this. @fl0wo have you found any workaround in the meantime?

I'm partially migrating from AWS to CF, at least for the resources it make sense to migrate from.

[...]

PR #354, opened 2024-11-29
refactor: use `node:events` instead of relative path 

vicb:
Change node:events import from relative to package import.

@pi0 There are a few relative imports for the net package. I left them untouched to stay on the safe side.
Let me know if it's the right thing to do - if it is, I'd like to add comment there too.
Thanks 🙏

Context: https://github.com/unjs/unenv/issues/353#issuecomment-2508064030

### DIFF
--- a/src/runtime/node/tls/internal/tls-socket.ts
+++ b/src/runtime/node/tls/internal/tls-socket.ts
@@ -1,5 +1,5 @@
 import type tls from "node:tls";
-import { Socket } from "../../net";
+import { Socket } from "node:net";
 import { createNotImplementedError } from "../../../_internal/utils";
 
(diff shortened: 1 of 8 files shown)


src/runtime/node/tls/internal/tls-socket.ts
```diff
@@ -1,5 +1,5 @@
 import type tls from "node:tls";
-import { Socket } from "node:net";
+import { Socket } from "../../net";
```
pi0, line 2: Thanks for fix/revert, it might be also needed as part of doing #323 👍

pi0 (approved):
thanks!

vicb:
Thank you!

PR #351, opened 2024-11-25
refactor(cloudflare): reimplement `module:createRequire` for latest workerd

vicb:
createRequire was implemented in v1.20240909
See https://github.com/cloudflare/workerd/releases/tag/v1.20240909.0

PR #356, opened 2024-12-02
fix(cloudflare): import notImplemented using a relative path

vicb:
This was trigering the following error:

```
-> % npx wrangler dev

 ⛅️ wrangler 3.91.0
-------------------


✘ [ERROR] Build failed with 1 error:

  ✘ [ERROR] Could not resolve "src/runtime/_internal/utils"
  
  
  node_modules/.pnpm/unenv@file+..+unenv/node_modules/unenv/runtime/node/module/$cloudflare.mjs:53:31:
        53 │ import { notImplemented } from "src/runtime/_internal/utils";
           ╵                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  
    You can mark the path "src/runtime/_internal/utils" as external to exclude it from the bundle,
  which will remove this error.
  
```

edits: 
- the cumprit is this [PR](https://github.com/unjs/unenv/pull/351) last week by me
- I create https://github.com/unjs/unenv/issues/357 to try to solve this at build time

### DIFF
--- a/src/runtime/node/module/$cloudflare.ts
+++ b/src/runtime/node/module/$cloudflare.ts
@@ -54,7 +54,7 @@ import {
   syncBuiltinESMExports,
   wrap,
 } from "./index";
-import { notImplemented } from "src/runtime/_internal/utils";
+import { notImplemented } from "../../_internal/utils";
 
 const workerdModule = process.getBuiltinModule("node:module");
 


petebacondarwin (approved):
Is this the only case of this kind of import?

I assume that there is some tsconfig.json setting to make `src` based imports work within this repo? (Perhaps `baseUrl` - https://github.com/unjs/unenv/blob/main/tsconfig.json#L3?)

If there is stuff like that going on in the repo, I would expect there to be some transpilation happening when the package is compiled to JS that would update these imports to work in vanilla node? Why didn't that happen? Perhaps an alternative fix would be to ensure that unenv's build step fixes these imports automatically?

vicb:
> Is this the only case of this kind of import?

yes - I amended the description, you might not have seen it, that's something I introduced last week.
Existing import are relative.

> I assume that there is some tsconfig.json setting to make src based imports work within this repo? (Perhaps baseUrl - https://github.com/unjs/unenv/blob/main/tsconfig.json#L3?)

> If there is stuff like that going on in the repo, I would expect there to be some transpilation happening when the package is compiled to JS that would update these imports to work in vanilla node? Why didn't that happen? Perhaps an alternative fix would be to ensure that unenv's build step fixes these imports automatically?

As of before this PR, the transpiled code (in `dist/`) would contain `import { notImplemented } from "src/runtime/_internal/utils";` which is not a valid path.

It should be possible to detect that at build time, I'll create an issue. For now, let's merge the fix if it looks ok.

petebacondarwin:
> It should be possible to detect that at build time, I'll create an issue. For now, let's merge the fix if it looks ok.
👍

PR #359, opened 2024-12-02
chore: disallow `src/` imports with eslint

### LINKED ISSUE #357: build: detect invalid path at build time
### Describe the feature

Ref: https://github.com/unjs/unenv/pull/356

I introduced `import { notImplemented } from "src/runtime/_internal/utils";` in #351 - the path was an autocomplete in VSCode.

The issue here is that `src/runtime/_internal/utils` which is not a valid path in dist.

It would be nice to detect that at build time

/cc @petebacondarwin 

### Additional information

- [x] Would you be willing to help implement this feature?

vicb:
fixes #357

```
-> % pnpm lint

> unenv@1.9.0 lint /Users/vberchet/code/work/unenv
> eslint . && prettier -c src test

/Users/vberchet/code/work/unenv/src/runtime/node/fs/index.ts
  16:1  error  'src/runtime/_internal/utils' import is restricted from being used by a pattern. Use relative imports instead  no-restricted-imports

✖ 1 problem (1 error, 0 warnings)

 ELIFECYCLE  Command failed with exit code 1.
 ```

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -26,6 +26,17 @@ export default unjs(
       "@typescript-eslint/no-unused-expressions": 0,
       "unicorn/prefer-global-this": 0,
       "unicorn/prefer-math-min-max": 0,
+      "no-restricted-imports": [
+        "error",
+        {
+          patterns: [
+            {
+              group: ["src/*"],
+              message: "Use relative imports instead.",
+            },
+          ],
+        },
+      ],
     },
   },
   {


pi0:
Just pushed https://github.com/unjs/unenv/commit/cb8fa54cdc272cbbdae7b9386863bf8bf17642c6 that does the same effect and fails build!

ESLint additionally is also good can you please revert failing file?

vicb:
> Just pushed https://github.com/unjs/unenv/commit/cb8fa54cdc272cbbdae7b9386863bf8bf17642c6 that does the same effect and fails build!

Thanks!

> ESLint additionally is also good can you please revert failing file?

Oops, good catch and sorry for that

PR #372, opened 2024-12-11
feat: `defineEnv` util

pi0:
This PR adds a new `defineEnv({ presets: [], ... }): { env }` utility aiming to replace `env(presets[]): env` util for integration usage.

The benefit of this one is that it can be extendable with more features, options and shortcuts.

Example:

```ts
const { env } = defineEnv({
 nodeCompat: true,
 presets: [myPreset],
 overrides: {}
});
```

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -1,5 +1,44 @@
-import type { Preset, Environment } from "./types";
+import type { Preset, Environment, CreateEnvOptions } from "./types";
 
+import nodeCompatPreset from "./presets/nodeless";
+
+/**
+ * Configure a target environment.
+ *
+ * @example
+ * ```ts
+ * const { env } = defineEnv({
+ *  nodeCompat: true,
+ *  presets: [myPreset],
+ *  overrides: {}
+ * });
+ */
+export function defineEnv(opts: CreateEnvOptions = {}): {
+  env: Environment;
+} {
+  const presets: Preset[] = [];
+
+  if (opts.nodeCompat) {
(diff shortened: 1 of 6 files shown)


pi0:
(merging as experimental to iterate and improve API features)

src/env.ts
```diff
@@ -1,5 +1,44 @@
...
+    presets.push(...opts.presets);
+  }
+
+  if (opts.overrides) {
+    presets.push(opts.overrides);
+  }
+
+  const resolvedEnv = env(...presets);
+
+  return { env: resolvedEnv };
```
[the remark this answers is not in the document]
pi0: Both for extendability (you see in next PR!)

src/types.ts
```diff
@@ -1,3 +1,20 @@
+export interface CreateEnvOptions {
+  /**
+   * Enable Node.js compatibility (nodeless) preset.
```
vicb, line 3: Explicitly add the default value

pi0: I think **Enable ...** implies that default is false won't it?

pi0: https://github.com/unjs/unenv/pull/373/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR5

src/env.ts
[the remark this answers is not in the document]
pi0: - #373 updates resolvedEnv before return
- #374 exposes presets alongside env

PR #373, opened 2024-12-11
feat(defineEnv): support resolving paths

pi0:
This PR adds a feature to new `defineEnv` util (introduced in #372) to allow resolving env config with absolute resolved paths.

ESM resolves are made using [mlly](https://github.com/unjs/mlly) (using the same Node.js ESM resolution logic with CJS compat and more features)

Resolve paths can be configured to extended. by default, a resolution from `unenv` itself will be done as last fallback.

~~Pending #374 to allow external presets paths.~~ (updates tiny change on #374 to use `meta.url` with ESM compat for URL)


src/env.ts
```diff
@@ -32,6 +32,35 @@ export function defineEnv(opts: CreateEnvOptions = {}): {
 
   const resolvedEnv = env(...presets);
 
+  if (opts.resolve) {
+    const resolveOpts: ResolveOptions = {
+      url: [
+        ...(opts.resolve === true ? [] : opts.resolve.paths || []),
```
vicb: nit
```suggestion
        ...(opts.resolve === true ? [] : opts.resolve.paths ?? []),
```

vicb (approved):

src/env.ts
pi0: nullish is invalid (like `0` or `false`) it will cause runtime error when spreading in same line.

pi0:
One thing left, we need to add `unenv/` itself to aliases (+ preset names) but delaying it to see if we need.

PR #374, opened 2024-12-11
feat: presets meta

pi0:
This PR adds `meta` to the presets which is useful for external presets meta. 

Presets are exposed from `defineEnv` for inspection and `path` is useful for #373

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -15,6 +15,7 @@ import nodeCompatPreset from "./presets/nodeless";
  */
 export function defineEnv(opts: CreateEnvOptions = {}): {
   env: Environment;
+  presets: Preset[];
 } {
   const presets: Preset[] = [];
 
@@ -32,7 +33,7 @@ export function defineEnv(opts: CreateEnvOptions = {}): {
 
   const resolvedEnv = env(...presets);
 
-  return { env: resolvedEnv };
+  return { env: resolvedEnv, presets };
 }
 
 /**
--- a/src/presets/nodeless.ts
+++ b/src/presets/nodeless.ts
@@ -1,7 +1,13 @@
 import { NodeBuiltinModules, mapArrToVal } from "../utils";
 import type { Preset } from "../types";
+import { version } from "../../package.json";
 
 const nodeless: Preset & { alias: Map<string, string> } = {
+  meta: {
+    name: "unenv:nodeless",
+    version,
+  },
+
   alias: {
     // Generic mock for built-ins
     ...mapArrToVal("unenv/runtime/mock/proxy-cjs", NodeBuiltinModules),
--- a/src/types.ts
+++ b/src/types.ts
@@ -27,17 +27,17 @@ export interface Preset {
     /**
      * Preset name.
      */
-    name?: string;
+    readonly name?: string;
 
     /**
      * Preset version.
      */
-    version?: string;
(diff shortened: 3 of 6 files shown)


src/env.ts
```diff
@@ -15,6 +15,7 @@ import nodeCompatPreset from "./presets/nodeless";
  */
 export function defineEnv(opts: CreateEnvOptions = {}): {
   env: Environment;
+  presets: Preset[];
```
vicb, line 18: nit: update l10 above with `presets`

src/presets/nodeless.ts
```diff
@@ -1,7 +1,13 @@
 import { NodeBuiltinModules, mapArrToVal } from "../utils";
 import type { Preset } from "../types";
+import { version } from "../../package.json";
 
 const nodeless: Preset & { alias: Map<string, string> } = {
+  meta: {
+    name: "unenv:nodeless",
```
vicb, line 7: question: this is referred to as `nodeCompat` in `defineEnv` - do we want to unify the meta and file names here too?

src/types.ts
```diff
@@ -23,6 +23,23 @@ export interface Environment {
 }
 
 export interface Preset {
+  meta?: {
```
vicb, line 26: do we want name/version (at least) as read-only?

src/presets/nodeless.ts
```diff
@@ -1,7 +1,13 @@
 import { NodeBuiltinModules, mapArrToVal } from "../utils";
 import type { Preset } from "../types";
+import { version } from "../../package.json";
 
 const nodeless: Preset & { alias: Map<string, string> } = {
```
vicb, line 5:
```suggestion
const nodeless: Preset = {
```

src/env.ts
```diff
@@ -32,7 +33,7 @@ export function defineEnv(opts: CreateEnvOptions = {}): {
 
   const resolvedEnv = env(...presets);
 
-  return { env: resolvedEnv };
+  return { env: resolvedEnv, presets };
```
vicb, line 36: What would the usage look like?

If we usually need both env and preset maybe:
`const presets = definePreset(opts: CreateEnvOptions = {})` and `env(...presets)` would work too.

_Mostly thinking out loud here_

vicb (approved):
LGTM with a few minor comments/questions

src/presets/nodeless.ts
pi0: Yes i plan to do via followup PRs to constantly use `nodeCompat` term. But it requires wrangler first to adopt so we won't break it..

src/env.ts
pi0: `const { env } = defineEnv(opts)`. `env` is current "builder config" but we can add more utils, meta, etc top level without breaking this. having all on same level might be tricky to maintain long term.

src/env.ts
pi0: Not sure to undrestand, L10 is part of example code:

image

src/types.ts
pi0: actually whole preset shoudln't be mutated.. marked new keys as readonly.

src/presets/nodeless.ts
pi0: lets do it in future PRs.

PR #376, opened 2024-12-16
feat(cloudflare): use workerd implementation of node:dns

vicb:
Implemented in https://github.com/cloudflare/workerd/releases/tag/v1.20241213.0

/cc @anonrig

### DIFF
--- a/src/presets/cloudflare.ts
+++ b/src/presets/cloudflare.ts
@@ -15,6 +15,7 @@ const cloudflareNodeCompatModules = [
   "assert/strict",
   "buffer",
   "diagnostics_channel",
+  "dns",
   "events",
   "path",
   "path/posix",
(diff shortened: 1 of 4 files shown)


src/presets/cloudflare.ts
```diff
@@ -15,6 +15,7 @@ const cloudflareNodeCompatModules = [
   "assert/strict",
   "buffer",
   "diagnostics_channel",
+  "dns",
```
anonrig, line 18: and dns/promises

vicb: Very good catch, thanks!

anonrig (approved):

pi0 (approved):

vicb:
Thanks for the reviews!

PR #378, opened 2024-12-18
fix(defineEnv): fallback support for `unenv-nightly`

pi0:
This PR is a hotfix for `defineEnv` utility when it is running in an aliased `unenv-nightly` package and when package manager (pnpm) does **not** rename `"name"` field in `package.json`, it causes ESM resolution to fail.

Workaroud is to fallback to `unenv/` => `unenv-nightly/` and then id as-is.

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -44,7 +44,22 @@ export function defineEnv(opts: CreateEnvOptions = {}): {
     const resolveOpts: ResolveOptions = {
       url: resolvePaths,
     };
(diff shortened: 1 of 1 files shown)


vicb (approved):

vicb:
Thanks a lot ❤️

vicb:
@pi0 Can you think of a test that would fail reliably to test this? (I'd like to add this to wrangler)

Issue #380, opened 2025-01-07
Node: process.std{in,out,err} do not implement ReadStream/WriteStream


gpanders:
### Environment

Cloudflare Workers

### Reproduction

Check `process.stdout.isTTY` and see that it is not a boolean (it is a function).

### Describe the bug

`process.stdout` is [mocked](https://github.com/unjs/unenv/blob/5d6aec38968012e78d2c0cda285e74b0f5f31ec3/src/runtime/node/process/internal/process.ts#L194-L196), so `process.stdout.isTTY` returns a function (the mock function), not a boolean. Likewise for `process.stdin.isTTY` and `process.stderr.isTTY`.

This means that well-behaved code that wants to check if stdout is a TTY (e.g. for emitting ANSI escape codes for colored logs) cannot do so in a portable way, since `if (process.stdout.isTTY)` will be true when polyfilled by unenv.

`process.stdout` and `process.stderr` should implement the [`WriteStream`](https://github.com/unjs/unenv/blob/5d6aec38968012e78d2c0cda285e74b0f5f31ec3/src/runtime/node/tty/internal/write-stream.ts) polyfill and `process.stdin` should implement the [`ReadStream`](https://github.com/unjs/unenv/blob/5d6aec38968012e78d2c0cda285e74b0f5f31ec3/src/runtime/node/tty/internal/read-stream.ts) polyfill. This allows portable code to work as expected in all environments.

pi0:
Nice find. Would you like to open a PR to fix?

PR #381, opened 2025-01-07
fix(node:process): implement `stdout`, `stderr` and `stdin` with `node:tty`

gpanders:
Resolves: https://github.com/unjs/unenv/issues/380

### DIFF
--- a/src/runtime/node/process/internal/process.ts
+++ b/src/runtime/node/process/internal/process.ts
@@ -1,5 +1,6 @@
 // Source: https://github.com/defunctzombie/node-process/blob/77caa43cdaee4ea710aa14d11cea1705293c0ef3/browser.js
 import type nodeProcess from "node:process";
+import { ReadStream, WriteStream } from "node:tty";
 import mock from "../../../mock/proxy";
 import empty from "../../../mock/empty";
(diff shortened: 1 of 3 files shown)


src/runtime/node/process/internal/process.ts
```diff
@@ -3,6 +3,7 @@ import type nodeProcess from "node:process";
 import mock from "../../../mock/proxy";
 import empty from "../../../mock/empty";
 import { notImplemented } from "../../../_internal/utils";
+import { ReadStream, WriteStream } from "../../tty/index.mjs";
```
pi0, line 6: Here, we can use `node:tty`. In pure unenv mode, it will use exact same polyfills (via alias) and in hybrid mode, it might leverage runtime (for example workerd) natives.

pi0:
```suggestion
import { ReadStream, WriteStream } from "node:tty";
```

gpanders:
@pi0 Do you need to manually restart CI?

pi0 (approved):

pi0:
@gpanders can you please check type error? It seems we might need to assign some fd to it

pi0 (approved):

pi0:
Thanks!

PR #392, opened 2025-01-20
refactor!: update `runtime/node` structure

pi0:
This PR makes a major refactor on `runtime/node/*` structure to avoid `/index` entries. /cc @vicb

This makes node structure closer to Node.js [lib dir](https://github.com/nodejs/node/tree/main/lib) ([preview](https://github.com/unjs/unenv/tree/refactor/node-no-index/src/runtime/node))

Since [`@cloudflare/unenv-preset`](https://npmjs.com/@cloudflare/unenv-preset/) is now fully externalized, we can safety remove entries.


vicb:
> This PR makes a major refactor on `runtime/node/*` structure to avoid `/index` entries. /cc @vicb
> 
> This makes node structure closer to Node.js [lib dir](https://github.com/nodejs/node/tree/main/lib).
> 
> Since [`@cloudflare/unenv-preset`](https://npmjs.com/@cloudflare/unenv-preset/) is now fully externalized, we can safety remove entries.

Hey Pooya,

Can this put this PR on hold for a few days?

I think @anonrig is going to submit a PR to use the builtin implementation for `node:net` and `node:timers`. It would be nice to have a RC with the current structure as we are still using it in `wrangler`.

On our side, we should figure out a way to solve the issue that got the use of `@cloudflare/unenv-preset` [reverted](https://github.com/cloudflare/workers-sdk/pull/7804).

If it's possible, it would be nice to allow a few days to allow to figure things out and investigate the impact this PR might have.

Thanks

/cc @anonrig @IgorMinar @petebacondarwin

pi0:
Sure I can wait for a few days till you can adopt rc releases back.

The impact of this PR is that you have to remove `/index` parts in your preset.

Regarding the Nitro (Nuxt) dep conflict issue, the solution is to stop using nightly npm aliases I think it should be the highest priority.

vicb:
> Sure I can wait for a few days till you can adopt rc.0 back. I like to do next rc with this changes.

Thanks.
First step is for us to migrate to [rc.0](https://github.com/cloudflare/workers-sdk/pull/7806) - that should land soon.
But we disabled the unenv module resolution that was causing a regression (and reverted to using `require.resolvce()`)

> Regarding the Nitro (Nuxt) dep conflict issue, the solution is to stop using nightly npm aliases I think it should be the highest priority.

As mentioned, we are migrating to rc.0. But I'd like to get to the bottom of the issue so that we do not re-introduce the regression at some point.

src/presets/index.ts
```diff
@@ -1,3 +1,2 @@
 export { default as node } from "./node";
 export { default as nodeless } from "./nodeless";
-export { default as cloudflare } from "./cloudflare";
```
vicb, line 3: Hey @pi0, do you think it would be possible to keep the cloudflare preset here for one more RC and remove it in the next one? Maybe moving the file to `node/$cloudflare-<module>.ts`.

This might allow for a smoother transition?

pi0:
Checked with @petebacondarwin. based C3 tests seem are ready to give better confidence about migration to cloudflare preset.

Issue #394, opened 2025-01-21
Deno: Remove `process.env` polyfill


marvinhagemeister:
### Environment

Deno 2.1.6

### Reproduction

N/A

### Describe the bug

Since Deno 2 the `process` variable is available globally. There is code in `unenv` that polyfills it, which isn't needed.

https://github.com/unjs/unenv/blob/262814fdb6780daa3100b64a734d2123d7047d06/src/runtime/polyfill/deno-env.ts#L2

### Logs

```sh

```

PR #395, opened 2025-01-21
chore: remove old deno process polyfill

marvinhagemeister:
Since Deno 2 the `process` variable is available globally.

Fixes https://github.com/unjs/unenv/issues/394


pi0 (approved):
Thanks for PR. BTW we have dropped experimental presets for v2, it was a leftover file.

PR #396, opened 2025-01-21
feat(cloudflare): add `node:net` and `node:timers`

anonrig:
Adds node:net and node:timers since they're both fully implemented in workers right now.

### DIFF
--- a/src/presets/cloudflare.ts
+++ b/src/presets/cloudflare.ts
@@ -4,7 +4,7 @@ import { version } from "../../package.json";
 // Built-in APIs provided by workerd.
 // https://developers.cloudflare.com/workers/runtime-apis/nodejs/
 // https://github.com/cloudflare/workerd/tree/main/src/node
-// Last checked: 2024-10-22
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -60,7 +60,7 @@
     "typescript": "^5.7.3",
     "unbuild": "^3.3.1",
     "vitest": "^3.0.2",
-    "workerd": "^1.20250109.0",
+    "workerd": "^1.20250121.0",
```
vicb, line 63: node:timers is buggy in this release, let's wait for the next one

pi0 (approved):
Thanks. Is it already in workerd prod version?

vicb (changes requested):
As commented online, we should wait for the fixes to be merged. 

Could you please also add tests? 
And send the same changes to workers-sdk/packages/unenv-preset?

Thanks !

anonrig:
Workers prod now includes the necessary changes. We can land this pull-request.

PR #397, opened 2025-01-23
fix: add ./runtime/node/string_decoder/index/ to exports

KonghaYao:
fix most exports problem in nuxt , cloudflare and netlify 
https://github.com/nitrojs/nitro/issues/2631
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,11 @@
   "license": "MIT",
   "sideEffects": false,
   "exports": {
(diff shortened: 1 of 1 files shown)


pi0:
I appreciate the PR. 

Right way is to add them to [preset aliases](https://github.com/unjs/unenv/blob/main/src/presets/nodeless.ts#L79) can you please open a new PR with this? 🙏🏼

PR #399, opened 2025-01-24
fix: add `string_decoder` / `node:string_decoder` to nodeless aliases

KonghaYao:
I think string_decorder is a standard nodejs libraray, so I just add `string_decorder` to list. 
I'm not quite sure about the logic of the code in this part. If there's any issue with my code, you can make the necessary modifications. Thank you very much.

### DIFF
--- a/src/presets/nodeless.ts
+++ b/src/presets/nodeless.ts
@@ -65,6 +65,7 @@ const nodeless: Preset & { alias: Map<string, string> } = {
         "wasi",
         "worker_threads",
         "zlib",
+        "string_decoder",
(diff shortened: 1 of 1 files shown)


pi0:
It is already in L54 (https://github.com/KonghaYao/unenv/blob/main/src/presets/nodeless.ts#L54)

Current issues in Nitro (and Nuxt) are because we are still stuck with unenv v1. 

There are https://github.com/nitrojs/nitro/issues/1821 and https://github.com/nitrojs/nitro/issues/2631 issues to track it there (I will try to check also in weekend as it is a common issue, we should fix it sooner!)

PR #400, opened 2025-01-24
fix(cloudflare): node:net/promises does not exist

vicb:
Discussed with @anonrig.

@pi0 would it be possible to have a release after this fix is merged.

### DIFF
--- a/src/presets/cloudflare.ts
+++ b/src/presets/cloudflare.ts
@@ -19,7 +19,6 @@ const cloudflareNodeCompatModules = [
   "dns/promises",
   "events",
   "net",
-  "net/promises",
   "path",
   "path/posix",
   "path/win32",


anonrig (approved):

pi0:
This PR should be available under `unenv-nightly@2.0.0-20250124-174216-aabfecf`

pi0:
@vicb i can do one more rc (although really it wasn't the plan) but can you please at least kindly double check if latest nightly is fully working on your side? (so rc.2 would not be blocked anymore until you reflect #392 which is fairly easy)

vicb:
The problem with nightly is resolution :(

@anonrig could you please double check the nightly, just run `pnpm -F wrangler unenv@npm:unenv-nightly@2.0.0-20250124-174216-aabfecf`, rebuild wrangler `pnpm -F wrangler build` and then run your tests using `cli.js`. Please then report here.

On my side I'll take a quick look at #392 over the weekend but we should not block on cf as we are not using the `@cf/preset` for now anyway. So we have time sync before we switch back to that.

Does that sound good.

pi0:
What is a blocker for `@cf/preset` adoption? (is there a tracker?)

pi0:
https://www.npmjs.com/package/unenv/v/2.0.0-rc.1

anonrig:
> The problem with nightly is resolution :(
> 
> @anonrig could you please double check the nightly, just run `pnpm -F wrangler unenv@npm:unenv-nightly@2.0.0-20250124-174216-aabfecf`, rebuild wrangler `pnpm -F wrangler build` and then run your tests using `cli.js`. Please then report here.
> 
> On my side I'll take a quick look at #392 over the weekend but we should not block on cf as we are not using the `@cf/preset` for now anyway. So we have time sync before we switch back to that.
> 
> Does that sound good?

No issues found.

PR #415, opened 2025-02-06
fix(node:async_hooks): remove extra methods from `AsyncHook` class

### LINKED ISSUE #403: `async_hooks.AsyncHook` hooks has extra methods
### Environment

N/A

### Reproduction

https://github.com/unjs/unenv/blob/ada7c4093c69c9729f1b008a8ab7902389941624/src/runtime/node/async_hooks/internal/async-hook.ts#L25

AsyncHook class should not have functions that should be defined in async hook callbacks. Specifically any methods except enable and disable should be removed.

### Describe the bug

async_hook polyfill is wrong.

### Additional context

_No response_

### Logs

```sh

```

pi0:
resolves #403

### DIFF
--- a/src/runtime/node/internal/async_hooks/async-hook.ts
+++ b/src/runtime/node/internal/async_hooks/async-hook.ts
@@ -2,7 +2,9 @@ import type asyncHooks from "node:async_hooks";
 
 // https://nodejs.org/api/async_hooks.html
 
-class _AsyncHook implements asyncHooks.HookCallbacks {
+type NodeAsyncHook = ReturnType<typeof asyncHooks.createHook>;
+
+class _AsyncHook implements NodeAsyncHook {
   readonly __unenv__ = true;
 
   _enabled: boolean = false;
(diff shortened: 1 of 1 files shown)


anonrig (approved):

src/runtime/node/internal/async_hooks/async-hook.ts
```diff
@@ -22,34 +24,24 @@ class _AsyncHook implements asyncHooks.HookCallbacks {
     return this;
   }
 
-  init(asyncId: number, type: string, triggerAsyncId: number, resource: any) {
-    if (this._enabled && this._callbacks.init) {
-      this._callbacks.init(asyncId, type, triggerAsyncId, resource);
-    }
+  get [Symbol("init")]() {
```
vicb, line 27: Shouldn't that be `Symbol.for()`?

vicb: I looked into this:
Node uses `Symbol("...")` and has a ref in internal.

I think that having `[Symbol(...)]` makes the code unreachable? Is that what we want here? (maybe to satisfy the type).

pi0: Yeah same situation as Node (i wish they used `Symbol.for` we have so many issues with this always)

My goal was mainly semantic. it doesn't really matter for private symbols I guess.

PR #416, opened 2025-02-06
refactor: explicit imports with extension

pi0:
Followup from #392

(discovered from issues with node coverage)

In strict ESM resolution and with `node --experimental-strip-types`, extensions should be always provided and same for `/index`. 

This PR ~~does not~~ (seems issue with mkidst) should not affect any differences in dist output.


pi0:
could be conflicts need to redo again later -_-

src/runtime/node/internal/http/request.ts
```diff
@@ -1,7 +1,6 @@
 import type http from "node:http";
 import { Socket } from "node:net";
-// TODO: https://github.com/unjs/unenv/issues/365
-import { Readable } from "unenv/node/stream";
```
pi0, line 4: /cc @vicb two usages also updated  here

PR #417, opened 2025-02-06
build!: esm-only build

pi0:
This PR is the first attempt for converting unenv into esm-only package using [oxc](https://oxc.rs/) + [rolldown](https://rolldown.rs/) toolchain.

Runtime types are generated as isolated types which means more unknowns (instead of inferences) but named exports match and as we progress towards isolated-types support (there are 600+ issues we can't migrate immediately to that)

Runtime `.mjs` entries are now fully validated both for syntax issues (twice!) and also relative imports and exports are resolved with explicit extension and `/index` (if needed). mkdist was buggy in some cases but using AST-aware transform that's safer now.

Entry types are currently extracted as rolldown lacks isolated types extraction bundling. not a big deal will be an internal refactor later!

**breaking change:** I used this opportunity (as we are requiring valid ESM resolution) to use `unenv/*` instead of `unenv/runtime/*`. /cc @vicb if you like to consider this also in your preset upgrade PR


tada5hi:
won't this break (cjs or cjs+mjs) libraries which are built on top of unenv?

pi0:
@tada5hi since we don't enforce `import` condition, it normally shouldn't happen and rollup should handle the mixed situation, otherwise we add back stubs.

Going to integrate it into nitro v2, if you have other projects to evaluate, that would be really helpful 🙏🏼

PR #420, opened 2025-02-06
fix(node:crypto): fix `randomBytes` returning zero values

### LINKED ISSUE #277: crypto.randomBytes() always returns 0
### Environment

"unenv":  1.9.0
node: v20.11.0

### Reproduction

Polyfill a Nuxt app using

```
// nuxt.config.ts
import { env, nodeless } from 'unenv'
const { alias } = env(nodeless, {})

export default defineNuxtConfig({
  vite: {
    define: {
      global: 'globalThis',
    },
    resolve: {
      alias: {
        'node:buffer': alias['node:buffer'],
        crypto: alias['crypto'],
        stream: alias['stream'],
        http: alias['http'],
        https: alias['https'],
        zlib: alias['zlib'],
      },
    },
  },
})

```

```
// app.vue
import { randomBytes } from 'crypto'

const rand = randomBytes(32)
for(const num of rand) {
   console.log(num)
}

```

### Describe the bug

Uint8Array.slice in the following line does not update the bytes buffer, use [subArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/subarray) instead as changes to the new array will update the original array.

https://github.com/unjs/unenv/blob/main/src/runtime/node/crypto/internal/node.ts#L44
```
    getRandomValues(
      Uint8Array.prototype.slice.call(
        bytes,
        generated,
        generated + MAX_RANDOM_VALUE_BYTES,
      ),
    );
```

### Additional context

_No response_

### Logs

_No response_

pi0:
resolves #277 (for v2) 

thanks to @huseeiin for pointing out ❤️

### DIFF
--- a/src/runtime/node/internal/crypto/node.ts
+++ b/src/runtime/node/internal/crypto/node.ts
@@ -38,10 +38,9 @@ export const randomBytes: typeof nodeCrypto.randomBytes = (
     generated < size;
     generated += MAX_RANDOM_VALUE_BYTES
   ) {
-    // buffer.slice automatically checks if the end is past the end of
-    // the buffer so we don't have to here
     getRandomValues(
-      Uint8Array.prototype.slice.call(
+      // Use subarray to get a view of the buffer
+      Uint8Array.prototype.subarray.call(
         bytes,
         generated,
         generated + MAX_RANDOM_VALUE_BYTES,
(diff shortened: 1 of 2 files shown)


huseeiin:
how do i install v2?

PR #437, opened 2025-02-11
minor simplification of `unenvPreset`

vicb:


pi0:
Thanks! It was intentional for possibly adding logic such as #436

PR #438, opened 2025-02-11
refactor!: avoid proxy mock usage

pi0:
Proxy-based mocks are dangerous and can cause implicit behaviors in libraries.

Since we have migrated almost everything to explicit APIs for v2, this PR updates a few last places to use `notImplementedClass`. (this can cause some breaking changes however this way we can catch those usages and explicitly implement the desired behavior).

As part of migration implemented `http.Agent` polyfill (to polyfill `globalAgent` exports)

/cc @anonrig @vicb

### DIFF
--- a/src/runtime/node/https.ts
+++ b/src/runtime/node/https.ts
@@ -1,12 +1,13 @@
 // https://nodejs.org/api/https.html
 import type nodeHttps from "node:https";
 import { notImplemented, notImplementedClass } from "../_internal/utils.ts";
+import { Agent as HttpAgent } from "./internal/http/agent.ts";
 
 export const Server: typeof nodeHttps.Server =
   /*@__PURE__*/ notImplementedClass("https.Server");
 
 export const Agent: typeof nodeHttps.Agent =
-  /*@__PURE__*/ notImplementedClass("https.Agent");
+  HttpAgent as unknown as typeof nodeHttps.Agent;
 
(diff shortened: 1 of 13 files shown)


src/runtime/node/https.ts
```diff
@@ -1,13 +1,15 @@
...
 export const Server: typeof nodeHttps.Server =
   /*@__PURE__*/ notImplementedClass("https.Server");
 
-export const Agent: typeof nodeHttps.Agent = mock.__createMock__("https.Agent");
-export const globalAgent: typeof nodeHttps.globalAgent = new Agent();
+export const Agent: typeof nodeHttps.Agent =
+  /*@__PURE__*/ notImplementedClass("https.Agent");
+
+export const globalAgent: typeof nodeHttps.globalAgent =
+  /*@__PURE__*/ new Agent();
```
pi0, line 12: not sure about this

src/runtime/web/performance/_performance.ts
```diff
@@ -18,8 +17,8 @@ export class _Performance<
   _entries: PerformanceEntry[] = [];
   _resourceTimingBufferSize = 0;
 
-  navigation = mock.__createMock__("PerformanceNavigation");
-  timing = mock.__createMock__("PerformanceTiming");
+  navigation = undefined as any;
```
pi0, line 20: this is undefined in Node

PR #445, opened 2025-02-12
refactor!: remove extra node exports

pi0:
Remove extra node exports (that don't exist in `node:` modules but unenv polyfills were exporting).

Also adds some internals wrongly missed due to coverage script error.

/cc @vicb @anonrig 

[WIP] Added tests first. with

 `pnpm node-ts --watch ./test/node-coverage.ts`

### DIFF
--- a/src/runtime/node/module.ts
+++ b/src/runtime/node/module.ts
@@ -141,24 +141,24 @@ export const _readPackage = /*@__PURE__*/ notImplemented("module._readPackage");
 export const _pathCache = Object.create(null);
 export const globalPaths = ["node_modules"];
 
-export class _Module implements NodeJS.Module {
-  require: NodeJS.Require;
-  id: string;
-  filename: string;
-  exports = Object.create(null);
-  parent = undefined;
-  loaded = true;
-  children = [];
-  isPreloading: boolean = false;
-  path: string = "/";
-  paths: string[] = [];
+// export class _Module implements NodeJS.Module {
+//   require: NodeJS.Require;
+//   id: string;
+//   filename: string;
+//   exports = Object.create(null);
+//   parent = undefined;
+//   loaded = true;
+//   children = [];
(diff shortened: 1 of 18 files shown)


pi0:
Remaining extras:

```
node:constants             233/233   (extra: O_DIRECT, O_NOATIME, RSA_SSLV23_PADDING, and 137 more...)
node:crypto                68/68     (extra: createCipher, createDecipher, default.createCipher, and 1 more...)
node:dns/promises          46/46     (extra: ADDRCONFIG, ALL, V4MAPPED, and 3 more...)
node:process               88/88     (extra: process, default.throwDeprecation, default.mainModule, and 2 more...)
node:zlib                  40/40     (extra: default.Z_TREES, default.BROTLI_DECODE, default.BROTLI_ENCODE, and 66 more...)
```


Apart from `node:process` that needs further refactoring (to convert into proper class with named exports) other extras are exports from legacy Node versions wondering would it be worth to even keep them around? /cc @unjs/cloudflare

src/runtime/node/module.ts
```diff
@@ -130,28 +135,30 @@ export const _resolveFilename = /*@__PURE__*/ notImplemented(
...
-  path: string = "/";
-  paths: string[] = [];
-
-  constructor(id = "index.js") {
-    this.id = id;
-    this.filename = id;
-    this.require = createRequire(id);
-  }
-}
+// export class _Module implements NodeJS.Module {
```
pi0, line 144: @vicb i checked cloudflare preset also it seems it doesn't need it. by any chance do you remember why we initially added this export? (if no use we can cleanup code)

vicb: It was introduced in [June](https://github.com/unjs/unenv/pull/269/) before I joined cf.

Maybe @IgorMinar @petebacondarwin or @dario-piotrowicz can help?

PR #450, opened 2025-02-12
refactor!: rewrite `process` as class

pi0:
We have implemented `node:process` with named exports to mimic `global.process` and allow hybrid implementation interface however real process in Node.js extends EventEmitter (`process instanceof events.EventEmitter === true`) and we cannot have features such as lazy getters or warning on access with just named exports. [current process polyfill](https://github.com/unjs/unenv/blob/78d4520a1ed557950cd7fb4f6d884346a1cccfef/src/runtime/node/internal/process/process.ts) is also kinda messy!

with This PR:

- process extends `EventEmitter`
- `stdin`/`stdout`/`stderr` are lazily initialized for perf on first access (named exports from `node:process` deoptimizes!)
- Dummy getters clearly separated and dummy objects are empty object
- All optional props (that only exist in certain situations) + `_*` internals marked as `undefined` to avoid implicit behavior and also reduce bundle
- Internal exposes single `UnenvProcess` class accepting `env`, `hrtime` and `nextTick` implementation from constructor.
- Main entry handles named exports and provides implementation.
- `unenv/polyfill/process` extends `globalThis.process` (if exists) with a `Proxy` that only fallbacks to unenv one if does not has same prop (relevant to Nitro)

### hybrid presets

With this new implementation, hybrid presets such as clouflare can directly pass internal implementation:

```ts
import { Process as UnenvProcess } from "unenv/internal/process/process";

const process = new UnenvProcess({
  env: /* custom */,
  hrtime: /* custom */,
  nextTick: /* custom */,
});
```


pi0:
merging to validate with rc. @vicb if you have any feedback (or was thinking to revert this altogether, we can discuss, don't worry 👍🏼 )

PR #452, opened 2025-02-13
fix(process): refine types

vicb:


src/runtime/node/process.ts
```diff
@@ -73,6 +73,8 @@ export const {
   stdin,
   stdout,
   title,
+  throwDeprecation,
+  traceDeprecation,
```
pi0, line 77: This two are tricky. Check node coverage + `require('process').throwDeprecation` is `undefined` in latest node.

vicb:
Thanks for the review @pi0, I updated the PR

pi0 (approved):

PR #454, opened 2025-02-13
refactor(process)!: extract nextTick to a separate file

vicb:
This is needed so that cloudflare (i.e. any runtime with a `nextTick` impl) can drop the polyfill

### DIFF
--- a/src/runtime/node/internal/process/nexttick.ts
+++ b/src/runtime/node/internal/process/nexttick.ts
@@ -1,33 +1,3 @@
-// https://nodejs.org/api/process.html#processhrtime
-export const hrtime: NodeJS.Process["hrtime"] = Object.assign(
-  function hrtime(startTime?: [number, number] | undefined) {
-    const now = Date.now();
-    // millis to seconds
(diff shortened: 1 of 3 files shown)


pi0 (approved):
The effect is same (other export is tree shakable) but lets do it for clarity

src/runtime/node/internal/process/nexttick.ts
pi0, line 1: can we rename it to nexttick.ts

vicb:
> The effect is same (other export is tree shakable) but lets do it for clarity

At least it didn't tree shake when I tested

pi0:
Right! #455

vicb:
> [`7625ec5`](/unjs/unenv/commit/7625ec58951a2ee8dcb091c7d37b67d5ba92a1c3)

I tried something similar in [7625ec5](https://github.com/unjs/unenv/commit/7625ec58951a2ee8dcb091c7d37b67d5ba92a1c3), thanks for teaching me the correct way :)

PR #455, opened 2025-02-13
refactor: mark `process` internals as side-effect free

pi0:
fixing root cause of https://github.com/unjs/unenv/pull/454


src/runtime/node/internal/process/nexttick.ts
```diff
@@ -4,10 +4,10 @@ export const nextTick: NodeJS.Process["nextTick"] = globalThis.queueMicrotask
   ? (cb, ...args) => {
       globalThis.queueMicrotask(cb.bind(undefined, ...args));
     }
-  : _createNextTickWithTimeout();
+  : /*@__PURE__*/ createNextTickWithTimeout();
```
vicb, line 7: nice!
TIL, I tried something similar but misplaced the `/@__PURE__@/`.

PR #457, opened 2025-02-13
refactor(util): add inherits implementation

vicb:


pi0:
Ah missed it. Check #456 i think should fix it.

PR #465, opened 2025-02-18
feat: update node compat injects

pi0:
This PR improves node compat injects preset:

- Use `node:buffer` for `Buffer` inject (this way native aliases are considered) 
- Add `setImmediate` and `clearImmediate` from `node:timers`

Note: Test updated as currently `resolve: true` does not respects aliases fixing in next one!

/cc @vicb

### DIFF
--- a/src/preset.ts
+++ b/src/preset.ts
@@ -70,9 +70,11 @@ export const nodeCompatAliases = {
 } as const;
 
 export const nodeCompatInjects = {
-  global: "unenv/polyfill/globalthis",
   process: "unenv/node/process",
-  Buffer: ["unenv/node/buffer", "Buffer"],
+  global: "unenv/polyfill/globalthis",
+  Buffer: ["node:buffer", "Buffer"],
+  clearImmediate: ["node:timers", "clearImmediate"],
+  setImmediate: ["node:timers", "setImmediate"],
 } as const;
 
 export const npmShims = {
--- a/test/env.test.ts
+++ b/test/env.test.ts
@@ -42,10 +42,12 @@ describe("defineEnv", () => {
         expect(existsSync(path), path).toBe(true);
       }
       for (const inject of Object.values(env.inject)) {
-        expect(
-          existsSync(Array.isArray(inject) ? inject[0] : inject),
-          inject.toString(),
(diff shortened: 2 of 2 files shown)


vicb:
Thanks for ccing me.
We will have to set the timers inject to false for cloudflare.
I have been thinking if inheriting presets really makes sense but I'm not sure if other solutions would be better...

pi0:
Sure, probably for injects, it makes sense that you manually override them based on wrangler needs.

I don't expect to add any more items to node compact injects, the changes was based on full cooperation I made.

BTW look at [this](https://platform-node-compat.pi0.workers.dev/) (deployed to cloudflare with [flags](https://github.com/pi0/platform-node-compat/blob/main/wrangler.json)  `nodejs_compat` + `no_nodejs_compat_v2` + latest compact date suggested by @petebacondarwin), both globals are missing.

vicb:
> BTW look at [this](https://platform-node-compat.pi0.workers.dev/) (deployed to cloudflare with [flags](https://github.com/pi0/platform-node-compat/blob/main/wrangler.json) `nodejs_compat` + `no_nodejs_compat_v2` + latest compat date suggested by @petebacondarwin), both globals are missing (not really sure what is going on there..)

That are disabled by no_v2 in workerd

vicb:
But it looks like we are missing a few perf classes too... I need to add those

pi0:
> That is disabled by no_v2 in worked

Yeah... We discussed previously if there is a way to tell wrangler not to do node polyfills yet pass through `nodejs_compat_v2` to native `workerd` but it needs disabling the full wrangler build (and even if we/users do disable, currently there is a warning about mixing flags).

vicb:
> > That is disabled by no_v2 in worked
> 
> Yeah... We discussed previously if there is a way to tell wrangler not to do node polyfills yet pass through `nodejs_compat_v2` to native `workerd` but it needs disabling the full wrangler build (and even if we/users do disable, currently there is a warning about mixing flags).

I fully agree that this is very confusing, I'll create an issue tomorrow on workers-sdk.

FYI https://github.com/cloudflare/workers-sdk/pull/6832

PR #468, opened 2025-02-18
refactor: prefer `node:*/promises` for `promises` export

pi0:
Leftover follow-up from #460

This PR changes `promises` export from `node:<id>` modules to use `node:<id>/promises` default export instead of using local paths which increases ability to implement hybrid by only implementing `/promises` variant.

Updated `inspector/promises.ts` to remove circular dep issue.

/cc @vicb can you please check specially for `node:timers/promises` if won't make any potential trouble for you?


vicb:
timers(/promises) is now implemented natively by workerd

pi0:
Thanks for confirm i guess it should be safe then

PR #471, opened 2025-02-24
implement base64url in buffer by switching to js-base64

### LINKED ISSUE #462: implement base64url in Buffer
### Environment

latest node and unenv

### Reproduction

Buffer.from('hello').toString('base64url')

### Describe the bug

base64url is not implemented

### Additional context

base64url is just base64.replaceAll("+", "-").replaceAll("/", "_").replace("==", "")

there's https://github.com/feross/buffer/pull/314 but the author is not proving it

### Logs

```sh

```

huseeiin:
fixes #462


src/runtime/node/internal/buffer/base64.js
pi0, line 1: Can you refactor this to `.ts` for better type checking?

huseeiin: there's a .d.ts, so you already get type checking when you import it

pi0: i understand but most of other codebase is using ts (types inlined)

pi0:
@huseeiin any specific reason you haven't gone with Deno stdlib implementation as your original idea? checking, it seems this lib has extra polyfills for legacy env without atob support which we probably don't need that concern here.

huseeiin:
> @huseeiin any specific reason you haven't gone with Deno stdlib implementation as your original idea? checking, it seems this lib has extra polyfills for legacy env without atob support which we probably don't need that concern here.

i will use the deno stdlib version if you want, i support that idea too!

PR #478, opened 2025-03-07
refactor(node:performance): standalone implementation

pi0:
Initial implementation of `node:performance` was based on `unenv/web/performance` with the goal of only extending Node.js specific differences.

However in practice, this caused many loop deps between this two.


src/runtime/node/internal/perf_hooks/performance.ts
```diff
@@ -37,11 +33,123 @@ const nodeTiming = {
...
+// PerformanceEntry
+export class PerformanceEntry implements nodePerfHooks.PerformanceEntry {
+  readonly __unenv__ = true;
+  detail: any | undefined;
+  entryType = "event" as any; // abstract
+  name: string;
+  startTime: number;
+  constructor(name: string, options?: PerformanceMarkOptions) {
+    this.name = name;
+    this.startTime = options?.startTime || performance.now();
```
vicb, line 45: Should this fallbacks to Date.now and use globalThis.performance?.now() (same as Performance#now)

src/runtime/node/internal/perf_hooks/performance.ts
```diff
@@ -37,11 +33,123 @@ const nodeTiming = {
...
+}
+
+// PerformanceMark
+export const PerformanceMark: typeof nodePerfHooks.PerformanceMark = class PerformanceMark
+  extends PerformanceEntry
+  implements nodePerfHooks.PerformanceMark
 {
+  entryType = "mark" as const;
+
+  constructor() {
```
vicb, line 69: Is this constructor needed? 
If so should it takes ...args?

pi0: not needed but typescript is nagging (it was before)

src/runtime/node/internal/perf_hooks/performance.ts
pi0: https://github.com/unjs/unenv/pull/478/commits/c916adc68d20d27e0226e2cc3fd656e43a88fe96

pi0: or better applied same change of #478

PR #493, opened 2025-04-01
process: return Node verion

vicb:


src/runtime/node/internal/process/process.ts
```diff
@@ -5,6 +5,8 @@ import {
   createNotImplementedError,
 } from "../../../_internal/utils.ts";
 
+const NODE_VERSION = "22.14.0";
```
petebacondarwin, line 8: Might be worth adding the comment about this needing to be a reasonably new version in the code for future reference.

src/runtime/node/internal/process/process.ts
```diff
@@ -92,11 +94,11 @@ export class Process extends EventEmitter implements NodeJS.Process {
   ppid: number = 100;
 
   get version() {
-    return "";
+    return `v${NODE_VERSION}`;
```
petebacondarwin, line 97: I note that one can attempt to write to `process.version` in Node.js it just doesn't update:

```
> process.version
'v18.20.6'
> process.version = "xxx"
'xxx'
> process.version
'v18.20.6'
```

with this just being a getter is that going to break with `process.version = "xxx"`?

petebacondarwin (approved):

pi0:
Can you help to make an automated [script](https://github.com/unjs/unenv/tree/main/scripts) to sync [nvmrc](https://github.com/unjs/unenv/blob/main/.nvmrc) to a generated `.ts` file? 🙏🏼

src/runtime/node/internal/process/process.ts
pi0: This is same behavior of class getters

```js
// 1
c=new (class { get version(){return "1"} })();c.version=2;console.log(c.version) 
```

src/runtime/node/internal/process/process.ts
pi0: we have reference version we sync unenv with in nvmrc

vicb:
> Can you help to make an automated [script](https://github.com/unjs/unenv/tree/main/scripts) to sync [nvmrc](https://github.com/unjs/unenv/blob/main/.nvmrc) to a generated `.ts` file? 🙏🏼

Great idea!
I have added that to the `build.ts` script, let me know if this looks ok

pi0 (approved):
thnx!

PR #502, opened 2025-05-08
fix(node:http): add `Agent.options` property

petebacondarwin:
resolves https://github.com/cloudflare/workers-sdk/issues/9141

Reproduction: https://github.com/EqualMa/workers-node-http-agent-options

### DIFF
--- a/src/runtime/node/internal/http/agent.ts
+++ b/src/runtime/node/internal/http/agent.ts
@@ -9,8 +9,8 @@ export class Agent extends EventEmitter implements nodeHttp.Agent {
   readonly freeSockets = {};
   readonly sockets = {};
   readonly requests = {};
-  readonly options: nodeHttp.AgentOptions | undefined;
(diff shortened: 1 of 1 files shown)


src/runtime/node/internal/http/agent.ts
```diff
@@ -9,5 +9,10 @@ export class Agent extends EventEmitter implements nodeHttp.Agent {
   readonly freeSockets = {};
   readonly sockets = {};
   readonly requests = {};
+  readonly options: nodeHttp.AgentOptions | undefined;
+  constructor(opts?: nodeHttp.AgentOptions) {
```
pi0, line 13: Consdering #9141 requires `options` being a defined object, wouldn't it make sense we always default to an empty object too?

petebacondarwin: Good idea.

pi0 (approved):

PR #536, opened 2025-12-18
fix(async_hooks): preserve store during async callbacks

onmax:
Related: https://github.com/better-auth/better-auth/issues/6613

## Problem

`AsyncLocalStorage.run()` clears the store synchronously, breaking async callbacks. When an async function awaits, the store becomes `undefined`.

## Reproduction

```bash
git clone --depth 1 --filter=blob:none --sparse --branch repro/unenv-als https://github.com/onmax/repros.git
cd repros && git sparse-checkout set unenv-als-broken-minimal
cd unenv-als-broken-minimal && pnpm i && pnpm test
```

**Expected**: `{ value: "test" }`
**Actual**: `undefined`

## Verify fix

```bash
git sparse-checkout add unenv-als-fixed
cd ../unenv-als-fixed && pnpm i && pnpm test
```

The `-fixed` folder includes a pnpm patch with the fix.

PS: there is a pre-existing issue with the node-coverage.test.ts snapshot failure exists on main (Node version difference) - unrelated to this PR.

### DIFF
--- a/src/runtime/node/internal/async_hooks/async-local-storage.ts
+++ b/src/runtime/node/internal/async_hooks/async-local-storage.ts
@@ -30,21 +30,42 @@ class _AsyncLocalStorage<T> implements nodeAsyncHooks.AsyncLocalStorage<T> {
     callback: (...args: TArgs) => R,
     ...args: TArgs
   ): R {
+    const previousStore = this._currentStore;
     this._currentStore = store;
-    const res = callback(...args);
-    this._currentStore = undefined;
-    return res;
+    try {
+      const res = callback(...args);
+      if (res != null && typeof (res as any).then === "function") {
+        return Promise.resolve(res).finally(() => {
+          this._currentStore = previousStore;
+        }) as R;
+      }
+      this._currentStore = previousStore;
+      return res;
+    } catch (error_) {
+      this._currentStore = previousStore;
+      throw error_;
+    }
   }
 
   exit<R, TArgs extends any[]>(
     callback: (...args: TArgs) => R,
     ...args: TArgs
   ): R {
-    const _previousStore = this._currentStore;
+    const previousStore = this._currentStore;
     this._currentStore = undefined;
-    const res = callback(...args);
-    this._currentStore = _previousStore;
-    return res;
+    try {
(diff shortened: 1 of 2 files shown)


pi0:
AsyncContext / AsyncLocalStorage requires native implementation in runtime. That is why platforms like stackblitz in browser for example cannot polyfill it.

Current implementation is safe (as it does not leaks contexts outside of sync).

onmax:
This fixes better-auth/better-auth#6613 - current impl is broken for any async callback (store cleared immediately).

I understand the concurrent limitation. This fix works for:
- ✅ Single async operations
- ✅ Nested runs
- ❌ Concurrent `Promise.all([run(), run()])` on same instance

Would you prefer:
1. Keep fix + add JSDoc warning about concurrent runs?
2. Different approach?

pi0:
Cloudflare workers have native AsyncLocalStorage support; this polyfill should not be used in this env (but instead leveraging `nodeCompat` / `nodejs_compat`). Sadly as i mentioned there is _no_ way to implement it safely in JS alone.

src/runtime/node/internal/async_hooks/async-local-storage.ts
```diff
@@ -30,21 +30,42 @@ class _AsyncLocalStorage<T> implements nodeAsyncHooks.AsyncLocalStorage<T> {
...
+    const previousStore = this._currentStore;
     this._currentStore = store;
-    const res = callback(...args);
-    this._currentStore = undefined;
-    return res;
+    try {
+      const res = callback(...args);
+      if (res != null && typeof (res as any).then === "function") {
+        return Promise.resolve(res).finally(() => {
+          this._currentStore = previousStore;
```
pi0, line 39: This is dangerous when two promises resolve in parallel, as there is no concurrency lock.

pi0:
It seems some progress in https://github.com/better-auth/better-auth/issues/6613#issuecomment-3694961868

Really appreciate your work on PR ❤️ however it is not something safe to do in prod.

onmax:
Thanks for taking a look! Will redirect to better-auth to document the `nodejs_compat` requirement.

PR #542, opened 2026-03-23
fix: use processModule as receiver when falling back in process polyfill (#541)

### LINKED ISSUE #541: Potentially incorrect polyfill for `globalThis.process`
### Environment

unenv: v2.0.0-rc.24
node: 24.0.1
environment: cloudflare pages

### Reproduction

-

## The issue we encountered

We deploy a Nuxt project to cloudflare pages. The project builds successfully but fails to startup with the following error:
```
Error: Failed to publish your Function. Got error: Uncaught TypeError: Cannot read private member #t from an object whose class did not declare it.
 at chunks/nitro/nitro.mjs:1:45995 in get stdout
  at chunks/nitro/nitro.mjs:1:52042 in get
  at chunks/nitro/nitro.mjs:1:334848
```

After tracing the issue, we located it to be related to `unenv`'s polyfill of the `process` global. The `#t` in the minified version is the `#stdout` private member.

## Where I think it stems from

The `process` polyfill is returning a partial polyfill using a proxy if the `process` global exists:
https://github.com/unjs/unenv/blob/f89b7ccb5c05da70b946319783acf1fa1f113e22/src/runtime/polyfill/process.ts#L1-L17

In the `get` trap, If the `prop` does not exist in the proxied object, it falls back to `node:process`'s implementation, **but it uses the original proxied object as the `this` value, as passed in the third argument of `Reflect.get`**. I think that's incorrect as it's not certain `receiver` contains the dependencies needed by the method/getter of `processModule`.

The process internals have the following `std` implementations: https://github.com/unjs/unenv/blob/f89b7ccb5c05da70b946319783acf1fa1f113e22/src/runtime/node/internal/process/process.ts#L54-L70

## A solution that worked for us

Making the following replacement on line 12 of [the process polyfill](https://github.com/unjs/unenv/blob/f89b7ccb5c05da70b946319783acf1fa1f113e22/src/runtime/polyfill/process.ts) seems to fix the issue:
```diff
-         return Reflect.get(processModule, prop, receiver);
+         return Reflect.get(processModule, prop, processModule);
```

Happy to make a PR if the solution above is satisfactory!

### Additional context

_No response_

### Logs

```sh

```

guoyangzhen:
### 🔗 Linked issue

Closes #541

### 📝 Description

The `process` polyfill uses a Proxy to layer unenv's polyfill over the original process object. When a property doesn't exist on the proxy target, it falls back to `processModule` (unenv's polyfill implementation):

```ts
return Reflect.get(processModule, prop, receiver);
```

The problem: `receiver` is the Proxy wrapping the **original** process, not `processModule`. When `processModule` has getters that rely on private fields (like `#stdout` in the `std` implementations), `Reflect.get` invokes the getter with `this = receiver` — which is the wrong object and lacks the private fields.

This causes:
```
Uncaught TypeError: Cannot read private member #t from an object whose class did not declare it.
```

### Fix

Pass `processModule` as the receiver instead of `receiver`:

```diff
- return Reflect.get(processModule, prop, receiver);
+ return Reflect.get(processModule, prop, processModule);
```

This ensures getters/methods on `processModule` are invoked with `this = processModule`, which has the expected private fields.

### Credit

Solution identified by the issue reporter @unjs/unenv#541.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the property access behavior in the process polyfill to ensure proper context binding when accessing properties.

### DIFF
--- a/src/runtime/polyfill/process.ts
+++ b/src/runtime/polyfill/process.ts
@@ -9,7 +9,7 @@ globalThis.process = originalProcess
         if (Reflect.has(target, prop)) {
           return Reflect.get(target, prop, receiver);
         }
-        return Reflect.get(processModule, prop, receiver);
+        return Reflect.get(processModule, prop, processModule);
       },
     })
   : processModule;


dreitzner (commented):
LGTM.

Tested with a relatively big eCommerce website and nuxt 4 on cloudflare pages.

Before:

Image

After:

Image

PR #554, opened 2026-06-12
fix(runtime/npm/debug): support full DEBUG namespace syntax + extend delimiter

### LINKED ISSUE #546: Bad debug polyfill
### Environment

not important

### Reproduction

not important

### Describe the bug

see https://github.com/unjs/unenv/blob/f89b7ccb5c05da70b946319783acf1fa1f113e22/src/runtime/npm/debug.ts#L9

this only enables a debugger if the debug env either is `*` or starts with the namespace.

namespace wildcards are not supported. e.g. `worker:*`

and most importantly multiple namespaces are not supported. e.g. `worker1,worker2`

### Additional context

_No response_

### Logs

```sh

```

tsushanth:
## Summary

Closes #546.

The `debug` polyfill at `src/runtime/npm/debug.ts` enabled a debugger only when DEBUG was exactly `*` or `env.startsWith(namespace)`. That silently dropped two well-documented features of the upstream [`debug`](https://www.npmjs.com/package/debug) package:

- **Comma- or whitespace-separated namespace lists** — `DEBUG=worker1,worker2` enabled NEITHER `worker1` nor `worker2` (only the literal string `worker1,worker2` as a prefix).
- **Wildcard patterns** — `DEBUG=worker:*` did not match `worker:db` / `worker:queue` because the env didn't `startsWith` the child namespace.

And because `startsWith` is unanchored, `DEBUG=work` would also have falsely enabled the `worker` namespace (no boundary).

## Fix

Implement the upstream parsing algorithm. Split DEBUG on `,` / whitespace, treat a leading `-` as a negation, glob-escape regex metacharacters, promote `*` to `.*`, anchor with `^…$`. A namespace is enabled when it matches at least one include pattern and no exclude pattern.

```ts
for (const raw of env.split(/[\s,]+/)) {
  if (!raw) continue;
  const negated = raw.startsWith("-");
  const pattern = negated ? raw.slice(1) : raw;
  const escaped = pattern
    .replace(/[\\^$.+?()|[\]{}]/g, String.raw\`\\\$&\`)
    .replace(/\*/g, ".*");
  const re = new RegExp(\`^\${escaped}\$\`);
  (negated ? skips : names).push(re);
}
```

Compiled patterns are cached in a single slot keyed on the env string so the common case (DEBUG unchanged between log calls) does not reparse. Cache invalidation is `===` on the string — DEBUG flips rarely at runtime and that's sufficient.

Also fixed `extend(ns, delimiter)` which previously discarded the delimiter argument and concatenated raw:

```diff
-extend: (ns: string, _del?: string) => createDebug(namespace + ns),
+extend: (ns: string, delimiter: string = ":") =>
+  createDebug(\`\${namespace}\${delimiter}\${ns}\`),
```

So `createDebug("app").extend("worker")` now yields `app:worker` (matching upstream) instead of `appworker`. The delimiter parameter is honoured when supplied.

## Tests

New file `test/runtime-npm-debug.test.ts`. 12 tests covering:

- `*`, exact match, env unset (baseline)
- comma-separated namespaces
- whitespace-separated namespaces
- wildcard pattern (`worker:*`)
- anchor guard (`worker:*` does NOT match the bare `worker`)
- negation (`-worker:internal`)
- prefix-not-substring (`DEBUG=work` does NOT enable `worker`)
- env-change-detection between calls
- `extend` default delimiter (`:`)
- `extend` custom delimiter

Verified the new tests fail on `main` (7/12 fail) and pass on this branch (12/12).

## Test plan

- [x] `npx vitest run test/runtime-npm-debug.test.ts` — 12/12 pass
- [x] Verified failure-on-master by stashing the production change and re-running — 7/12 fail, confirming the tests are real regressions and not just covering already-correct behavior
- [x] `npm run lint` clean for the touched files (the 5 prettier warnings surfaced are pre-existing in files this PR does not touch)

## Out of scope

The full upstream `debug` API also handles `humanize`, color picking, and the `enable()`/`disable()` runtime control. This PR keeps the existing pass-through stubs for those — they're separately tracked and out of scope for the namespace-matching regression.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced DEBUG environment variable parsing with improved pattern matching for namespace filtering, including support for wildcards, exclusion patterns, and proper delimiter handling.

* **Tests**
  * Added comprehensive test coverage for debug namespace matching behavior.

### DIFF
--- a/src/runtime/npm/debug.ts
+++ b/src/runtime/npm/debug.ts
@@ -2,11 +2,65 @@
 
 import type { Debug, Debugger, Formatters } from "debug";
 
+// Cache the compiled patterns keyed on the DEBUG env value so we don't reparse
+// per log line. The cache is a single slot — DEBUG flips rarely at runtime and
+// strict equality on the env string is enough to invalidate when it does.
+let cachedEnv: string | undefined;
+let cachedPatterns: { names: RegExp[]; skips: RegExp[] } | undefined;
+
+function compileDebugEnv(
+  env: string | undefined,
+): { names: RegExp[]; skips: RegExp[] } | undefined {
+  if (env === cachedEnv) return cachedPatterns;
+  cachedEnv = env;
+  if (!env) {
+    cachedPatterns = undefined;
+    return undefined;
+  }
+
+  // The `debug` package splits DEBUG on `,` and whitespace, then treats a
+  // leading `-` as a negation. Each token is glob-style — `*` matches anything.
+  // Escape regex metacharacters first, then promote `*` to `.*` so wildcards
+  // like `worker:*` work, and anchor with `^…$` so e.g. `worker:abc` doesn't
+  // get matched by a pattern intended only for `worker`.
+  const names: RegExp[] = [];
+  const skips: RegExp[] = [];
+  for (const raw of env.split(/[\s,]+/)) {
+    if (!raw) continue;
+    const negated = raw.startsWith("-");
+    const pattern = negated ? raw.slice(1) : raw;
+    if (!pattern) continue;
+    const escaped = pattern
+      .replace(/[\\^$.+?()|[\]{}]/g, String.raw`\$&`)
+      .replace(/\*/g, ".*");
+    const re = new RegExp(`^${escaped}$`);
+    (negated ? skips : names).push(re);
+  }
+  cachedPatterns = { names, skips };
+  return cachedPatterns;
+}
+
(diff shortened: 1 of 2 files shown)


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

PR #557, opened 2026-06-13
fix(env): strip trailing slash from node: builtin specifiers (#555)

### LINKED ISSUE #555: resolvePath should strip trailing slashes from node: builtin specifiers
### Environment

I've seen this bug in both node and deno

node v25.8.1

deno 2.8.2 (stable, release, aarch64-apple-darwin)
v8 14.9.207.2-rusty
typescript 6.0.3

### Reproduction

See description...

## Summary

`resolvePath` currently preserves trailing slashes when a specifier resolves to a `node:` builtin.

That is a problem because normal package imports may legally use a trailing slash, but `node:` builtin specifiers must not.

For example, Node accepts:

```js
await import("punycode/")
```
but rejects:
```js
await import("node:punycode/")
```
with:
```
Error [ERR_UNKNOWN_BUILTIN_MODULE]: No such built-in module: node:punycode/
```
Real-world case
This showed up with punycode, which exists both as an npm package and as a Node builtin.
A dependency imported punycode/, and unenv converted that to the builtin form. The trailing slash was preserved, so the final specifier became node:punycode/, which is invalid in Node.
Reproduction
```js
const { env } = defineEnv({
  nodeCompat: false,
  resolve: true,
  overrides: {
    alias: {
      punycode: "node:punycode",
      foo: "punycode/",
    },
  },
});
```
```js
env.alias.foo;
```
Current result:
```js
"node:punycode/"
```
Expected result:
```js
"node:punycode"
```
Expected behavior
When resolvePath resolves a specifier to a builtin module, it should normalize a trailing slash away before returning the node: form.
This normalization should only apply to builtins. Regular non-builtin imports like my-module/ should remain unchanged.

### Additional context

_No response_

### Logs

```sh

```

tsushanth:
Closes #555.

When an alias chain feeds `_resolve` a value like `punycode/`, `pathe`'s `resolveAlias` rewrites it through the `punycode -> node:punycode` mapping and tacks the trailing slash back on, producing `node:punycode/`. That string then hits the `id.startsWith("node:")` short-circuit and was returned verbatim. Node rejects the resulting specifier with `ERR_UNKNOWN_BUILTIN_MODULE` because trailing slashes are not legal on `node:` builtins, even though the equivalent package form `punycode/` is fine.

The fix normalises only inside the `node:` branch and only when the trimmed name is a real builtin. So:

- `node:punycode/` -> `node:punycode` (the bug)
- `node:punycode` -> unchanged
- `node:not-a-real-builtin/` -> unchanged (no surprise rewrites of unknown values)
- non-`node:` strings -> unchanged path

This keeps the regular package-path behaviour the issue calls out (`my-module/` stays `my-module/`).

## Tests

Three new cases under `describe("resolvePath")` in `test/env.test.ts`:

- the reporter's repro: an alias chain that ends at `punycode/` resolves to `node:punycode`
- a literal `node:punycode` alias stays unchanged
- a `node:not-a-real-builtin/` alias is left alone

All three pass with the patch and fail (or never run) against `main`. The pre-existing `resolves all nodeCompat paths` failure on `main` (`unenv/node/inspector/promises` missing) is unrelated; confirmed by running the suite on `main` without these changes.

## Release Notes

* **Bug Fixes**
  * Fixed resolution of Node builtin module aliases with trailing slashes to prevent loading invalid or unknown builtin identifiers.

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -120,6 +120,16 @@ function resolvePaths(
       id = resolveAlias(id, env.alias);
     }
     if (id.startsWith("node:")) {
+      // `node:` builtin specifiers must not carry a trailing slash; Node throws
+      // ERR_UNKNOWN_BUILTIN_MODULE for `node:punycode/` even though the package
+      // form `punycode/` is legal. Strip the slash only when the rest is a
+      // real builtin, so unrelated `node:`-prefixed strings are untouched.
+      if (id.endsWith("/")) {
+        const trimmed = id.slice(0, -1);
+        if (builtinModules.includes(trimmed.slice("node:".length))) {
+          return trimmed;
+        }
+      }
       return id;
     }
     if (builtinModules.includes(id)) {
--- a/test/env.test.ts
+++ b/test/env.test.ts
@@ -67,5 +67,42 @@ describe("defineEnv", () => {
         expect(existsSync(to), inject.toString()).toBe(true);
       }
     });
+
+    it("strips trailing slash when an alias chains to a node: builtin", () => {
+      const { env } = defineEnv({
+        nodeCompat: false,
+        resolve: true,
+        overrides: {
+          alias: {
+            punycode: "node:punycode",
+            foo: "punycode/",
+          },
+        },
+      });
+      // `node:punycode/` would throw ERR_UNKNOWN_BUILTIN_MODULE at runtime.
+      expect(env.alias.foo).toBe("node:punycode");
+    });
+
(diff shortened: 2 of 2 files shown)


tsushanth:
Apologies @ashtonjurgens — your #556 was open about 6 hours before this and touches the exact same files (`src/env.ts` + `test/env.test.ts`). The `Fixes:` keyword wasn't picked up by the issue sidebar so I missed the parallel. Closing in favor of yours.

